PR #1, opened 2016-01-27
Emojis

keithpitt:
Migrating our emoji handling to the new fronted

- [x] Add support for skin tone modifiers


vendor/emojis
```diff
@@ -0,0 +1 @@
+Subproject commit 70d9e588e8c3f49bb116c03f0d84cd1ac6ab41e2
```
sj26: This is a little behind now I think.

sj26:
Looks pretty good to me. :+1:

PR #3, opened 2016-01-30
New repo layout

toolmantim:
The idea is to create a "UI Library" where components contain Javascript + any assets they need… make them a first class citizen, rather than just in a "shared" folder. That way we can have a styleguide for those shared UI components, and we know not to leak in Stores or any non-UI related logic in there too.

Components can mean specific bits of interface that might use GraphQL, stores etc and the UI library.

This also just means we can maintain, update, fix and extend the UI repo without having to have our head around all the rest of the code… it's in its own folder.

Github's diff doesn't give a great overview of what this layout would look like.


keithpitt:
Yeah, folder structure is something I don't think anyone in the community has set a standard on yet. My only problem with folders for components is that you end up with lots of `index.js` files, which makes for crappy a "fuzzy file finder" experience in my edit (since all files are called `index.js`). Having said that, I'm happy to try a new layout and see how it goes!

It would be great along with this change, to see some guidelines in the Readme as to how to figure out where to put a file. Essentially "files that do (x) and don't have (y) go into the (z) folder" etc.

Also I noticed some import paths will break because of this change, so you'll need to double check them (also in our legacy UI gear you'll have to change the paths there as well)

toolmantim:
Kewl, yeah updating things to refer to the new paths isn't too hard (I made it work). 

This is the current/old layout:

old-layout

This is the new layout:

new-layout

You only end up with an index file if there's assets that go along with the component, which I'm guessing is going to be the 5% case.

Happy to leave this for now… just wanted to jot down my thoughts for this new repo, because I felt we might have been inheriting some old habits from the old stuff.

PR #6, opened 2016-02-03
My Builds tweaks

toolmantim:
Hides the badge when it's zero:

my-builds-count

I've also reduced the vertical size of the dropdowns and removed the borders. Not 100% with these, but getting closer.


keithpitt:
:+1:

PR #15, opened 2016-03-05
UI Elements for Teams

keithpitt:
Extracted the UI elements for the new teams pages.


keithpitt:
Meh, I'll just do this in the teams branch. No real point pulling it out really.

PR #17, opened 2016-03-16
Allow forcing team pipeline removal

sj26:
Following up buildkite/buildkite#1154 with some frontend to make it work:

This looks pretty ugly, but we don't have a nicer confirmation thingo.

Thoughts?

### DIFF
--- a/app/components/team/Pipelines/index.js
+++ b/app/components/team/Pipelines/index.js
@@ -93,9 +93,10 @@ class Pipelines extends React.Component {
     }), { onFailure: this.handleMutationFailure });
   };
 
-  handleTeamPipelineRemove = (teamPipeline, callback) => {
+  handleTeamPipelineRemove = (teamPipeline, force, callback) => {
     Relay.Store.commitUpdate(new TeamPipelineDeleteMutation({
-      teamPipeline: teamPipeline
+      teamPipeline: teamPipeline,
+      force: force
     }), { onSuccess: () => callback(null), onFailure: (transaction) => callback(transaction.getError()) });
   };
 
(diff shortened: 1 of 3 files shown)


toolmantim:
A double dialog thingy is a bit blegh, but I understand why we need it.

Even if we keep it, is there no way we can detect it ahead of time, on page load, to avoid the double dialog most of the time?

If not, then this will do for now :+1:

toolmantim:
Another way to do this might be to have a nice dialog that has a spinner saying "Checking pipeline permissions…" which does the check before you hit an "Okay, remove already" button

sj26:
We detect it by testing if a pipeline is accessible after the team pipeline is removed and aborting the transaction unless forced, so we could do that but it would require building something new in the backend which is going to be a bit complicated. Thoughts? Good enough for now(tm)?

toolmantim:
Cool. If you reckon we should merge, go for it!

I'm happy to build the UI like below and then we can add the engineering fanciness after to make it happen?

After clicking remove… (buttons disabled)



Pipeline will no longer be accessible:



Pipeline will still be accessible:

toolmantim:
Perhaps next time we should quickly sketch up the UI before doing the backend changes, to prevent redo'ing stuff? Just thinking that it must be frustrating for you?

I think we should just merge this now perhaps and change it later if/when we get to it.

sj26:
Yeah, I think maybe ship what we've got here and improve from there. @keith?

keithpitt:
Yeah, although double confirm is a bit poopy, we know it'd be this way when we first started, although it's a bit better than someone accidentally loosing access to a pipeline! Let's merge it and we can make it betterer soon

toolmantim:
:+1: :rocket:

PR #18, opened 2016-03-23
Switch to using autosize

toolmantim:
This switches to [jackmoore/autosize](https://github.com/jackmoore/autosize) which is has no deps, is pretty small, and handles a bunch of neat stuff for us. http://www.jacklmoore.com/autosize/


toolmantim:
I forgot to mention, it turns out buildo/react-autosize-textarea isn't so great because it really gets in the way if you want access to the actual textarea, and doesn't buy you a lot when you can simply use `componentDidMount` / `componentDidUnmount` to call the `autosize` function

toolmantim:
Hmm… that's a weird build error! It's in the package.json

keithpitt:
To fix the build I think you need to run `npm shrinkwrap --dev` which will save the `autoresize` package in the `npm-shrinkwrap.json` file, which is the npm equivalent of `Gemfile.lock`

app/components/shared/FormMarkdownEditorField.js
```diff
@@ -141,7 +140,6 @@ class FormMarkdownEdtiorField extends React.Component {
     e.preventDefault();
 
     this.markdownEditor.bold();
-    autoresizeTextarea(this.textarea);
```
keithpitt: You gotta trigger a manual resize here if you can, since we've altered the text inside the textarea, there's a chance that it's bumped the content down onto another line (which would need to change the height of the textarea)

PR #20, opened 2016-03-30
Fix the flex on panels with buttons

toolmantim:
I'm not sure how this ever worked… but it def broke if the panel text was smaller than the whole width

### DIFF
--- a/app/components/shared/Panel/intro-with-button.js
+++ b/app/components/shared/Panel/intro-with-button.js
@@ -21,7 +21,7 @@ class IntroWithButton extends React.Component {
 
     if(button) {
       button = (
-        <div className="ml3 flex-no-shrink">
+        <div className="ml3 flex-none">
           {button}
         </div>
       );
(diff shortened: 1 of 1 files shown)


keithpitt:
I added https://github.com/buildkite/frontend/blob/master/app/css/flexbox.css#L1 to try and combat that - but maybe it didn't work in all cases?

toolmantim:
I saw! But bass has `flex-none` for just that and we use it for other things, so I just switched to that?

The problem wasn't that though, it was that the text didn't fill up the rest of the space when there wasn't much text. I just changed it to the standard bass way…

```html
<div className="flex">
  <div className="flex-auto">Text</div>
  <div className="flex-none">Button</div>
</div>
```

You could also use the [margin-auto trick](https://medium.com/@samserif/flexbox-s-best-kept-secret-bd3d892826b6#.ysifolszw) if you didn't care about the text actually filling up the space…

```html
<div className="flex">
  <div className="mr-auto">Text</div>
  <div className="flex-none">Button</div>
</div>
```

PR #23, opened 2016-03-31
Our own spinner! :sparkles:

toolmantim:
spinner


app/components/shared/Spinner.js
```diff
@@ -0,0 +1,66 @@
...
+            </g>
+          </svg>
+        </div>
+      </div>
+    );
+  }
+
+  _arcColor() {
+    if (this.props.color) {
+      return "#7EAF25"; // lime-green
```
toolmantim: Maybe there's a fancy way to read this from our CSS?

keithpitt: I had a quick look. The only way I can see to do it is to inject the vars via webpack here: https://github.com/MadLittleMods/postcss-css-variables#variables-default - but it's probably more hassle than it's worth. If we ever change the color, as we'll checking the visual style of the site, we'll spot it pretty quick I think and be like: "oh, we should change it!"

keithpitt:
Awesome! Did you want this added to the section loader as well? So instead of "Loading..." it'll show this spinner? I can add that for you now if yes!

toolmantim:
Another thing I noticed is the autocomplete icon is a bit jarring when it switches to a spinner. I think it shouldn't change from a search icon, but probably just spin in the dropdown itself as you type. People sometimes morph the search icon to a spinner icon, but it's a bit yuk

toolmantim:
Yay, for the loading!

toolmantim:
Also I realized this morning that in inline SVG we can use the css classes… no need for CSS var importing!

toolmantim:
I also bumped the speed to the autocomplete. If you have any lag to our servers, then another lag on top of that to wait to finish typing feels really slow. It’d be great if it felt as fast as chrome… basically do it on every keypress.

toolmantim:
@keithpitt this is good to go!!!

keithpitt:
:rocket:

PR #25, opened 2016-04-01
Fix menu badge on Firefox

toolmantim:
Before:

before

After:

after

I also made the badge class more declarative using jsx and functions, and less imperative/code style. I think for UI components trying to keep it as tags and props and not putting too much JS in there means that you can more easily see how it’s working at a glance.

Also: no more need for `Features.NewNav`! :)


keithpitt:
:shipit:

PR #26, opened 2016-04-05
Add type="button" so they don't end up as submit buttons

toolmantim:
By default a browser uses the first button in the form with `type=submit` or no type at all when you submit a form with the keyboard. So if there's another field in the form, such as a Changelog Title, and you hit enter… the bold button would be clicked.

This just corrects the type of the buttons so they don't trigger on form submit.


keithpitt:
:+1:

PR #30, opened 2016-05-12
Fire a connected event on pusher connected

toolmantim:
For buildkite/buildkite#1238


toolmantim:
This is good to merge

keithpitt:
Something weird is going on, because it looks like this change is already in master somehow? 😕

toolmantim:
@keithpitt really? https://github.com/buildkite/frontend/blob/master/app/stores/PusherStore.js

PR #36, opened 2016-07-22
Add Users to the top nav

toolmantim:
To help people see that they can add users, this adds Users to the top nav:

users

I'm not 100% sold on it here, but without a nav rejig I'm not sure what another solution is for now.


toolmantim:
It'd be nice to come up with a better solution…

PR #41, opened 2016-08-10
Added a RelayPreloader class to load teams and organization information

keithpitt:
The `RelayPreloader` class gives us the ability to bootstrap data from the server into Relay on page load - helping us get rid of the spinners!


keithpitt:
It's a bit finicky, as in you need to maintain a copy of the GraphQL query created by Relay - but I think that's ok for the speed boosts you get!

We don't have _that_ many Relay pages that need bootstrap, so I think this should be totes fine.

PR #49, opened 2016-09-07
Create new Agents list

ticky:
- [x] Initial tooling setup
- [x] Data model
  - [x] Agent List
  - [x] Agent View
  - [x] Agent Stop
  - [x] Agent Token View
- [x] View structure
  - [x] Agent List
  - [x] Agent View
  - [x] Agent Token View
  - [ ] ~~Agent Quick Start guides~~
- [x] A few colour treatments, I guess?

### DIFF
--- a/app/components/agent/Show.js
+++ b/app/components/agent/Show.js
@@ -5,6 +5,7 @@ import classNames from 'classnames';
 
 import Button from '../shared/Button';
 import Emojify from '../shared/Emojify';
+import FlashesStore from '../../stores/FlashesStore';
 import FriendlyTime from "../shared/FriendlyTime";
 import PageWithContainer from '../shared/PageWithContainer';
 import Panel from '../shared/Panel';
@@ -138,7 +139,9 @@ class AgentShow extends React.Component {
     this.setState({ stopping: false });
   };
 
-  handleMutationError = () => {
+  handleMutationError = (transaction) => {
+    FlashesStore.flash(FlashesStore.ERROR, transaction.getError());
+
     this.setState({ stopping: false });
   };
 
--- a/app/components/agent/Show.js
+++ b/app/components/agent/Show.js
@@ -140,7 +140,7 @@ class AgentShow extends React.Component {
                 Meta Data
               </div>
               <div className="left sm-col-9 p2">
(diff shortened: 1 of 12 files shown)


ticky:
Rebased atop `master`, with new ESLint conventions properly followed 👌🏼

app/components/shared/FriendlyTime.js
```diff
@@ -4,15 +4,21 @@ import friendlyRelativeTime from '../../lib/friendlyRelativeTime';
 
 class FriendlyTime extends React.Component {
   static propTypes = {
-    value: React.PropTypes.string.isRequired
+    value: React.PropTypes.string.isRequired,
+    capitalized: React.PropTypes.bool.isRequired
+  };
+
+  static defaultProps = {
+    capitalized: false
```
keithpitt: I think this should be defaulted to `true` since that's what it was to start with?

app/components/agent/Show.js
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  handleStopButtonClick = (evt) => {
+    evt.preventDefault();
+
+    this.setState({ stopping: true });
+
+    const mutation = new AgentStopMutation({
+      agent: this.props.agent,
+      graceful: true
```
keithpitt: After looking at the current behaviour in the Stop button we do a `graceful: false` so we should do the same here I think!

app/components/agent/Show.js
```diff
@@ -0,0 +1,267 @@
...
+          <FriendlyTime value={agent.connectedAt} capitalized={true} />
+          {agent.pingedAt && agent.connectionState === 'connected' &&
+            <span> (last check-in was {<FriendlyTime value={agent.pingedAt} />})</span>
+          }
+        </span>
+      ));
+    }
+
+    if (agent.connectionState === 'disconnected') {
+      extras.push(this.renderExtraItem('Disconnected', <FriendlyTime value={agent.disconnectedAt} />));
```
keithpitt: Should we be using `CONNECTION_STATE_LABELS` here?

ticky: the label for `connectionState === 'lost'` is different; it could use the same label list, but I kept the behaviour the same as it currently is!

PR #52, opened 2016-09-12
Improve naming of components

ticky:
Some components have rather generic names, this updates them with their category if they have one.

Possibly worth discussing; some existing components have a `displayName` attribute applied, which sort of goes for a namespace-ish approach - which is more desirable here?


keithpitt:
🚀

PR #54, opened 2016-09-13
👓 Update ESLint rules to be more strict and useful

ticky:
- Change ESLint config to use textual severity for ease of reading
- Enforce brace style, spacing within braces (curly, square and otherwise)
- Require ES6 arrow functions to wrap parameter definitions, and spaces around their precious arrows
- Disallow dangling commas in object definitions
- Warn about use of non-strict equality operators like "!=" and "==" (we use these a lot, worth checking if we should be!)
- Warn about identifiers which are too short; "i", "j" and "k" are *not* useful variable names - be specific 🎉!!!
- Be strict about indentation within both plain JS code and JSX
- Prefer "const," avoid "var" in favour of "let"
- Disallow some rather unsafe stuff
- Require one definition per assigned var, allow multiple var declarations if all are unassigned
- Warn about using number parsers without supplying radix
- Require semicolons
- Require braces around if statements, even if single-line
- Disallow unused propType declarations
- Require style props to be objects
- Update files to comply with all "error"-level rules


keithpitt:
Yes to all this!

Only thing that hurts my brain is dangling commas and `{ foo: "bar" }` to `{foo: "bar"}` but I think I'll get used to it :P

ticky:
Definitely happy to take feedback on those points; also happy to discuss trying to keep some similarity between our expected Ruby and JS styles!

toolmantim:
I'm cool with these, happy to follow that style! 👍👍👍

keithpitt:
<3

PR #68, opened 2016-09-26
Optimize preloading of Pipelines

keithpitt:
- [x] Preloads pipeline name and build status
- [x] Delay loading of graphs and metrics
- [x] Delay loading of teams


keithpitt:
Gonna do some testing on prod to see how it feels!

PR #71, opened 2016-09-28
Pipeline responsivity tweaks

ticky:
Work in progress; hides the graph and stats on small screens


keithpitt:
I just checked this branch out locally and it looks great! I'm gonna merge this in and push to production since it's already like a bajillion times better than what we currently have :)

PR #73, opened 2016-10-05
Replace unicode down triangles with real icons

toolmantim:
I went to fix this Firefox bug:

firefox-before

and then I accidentally fixed all the dropdown triangles to be nicer than unicode:

and-then

They also look nice on mobile now too:

mobile-triangles

### DIFF
--- a/app/components/shared/Icon/index.js
+++ b/app/components/shared/Icon/index.js
@@ -35,6 +35,8 @@ const pathNodes = (icon) => {
       return require("./github").default;
     case 'twitter':
       return require("./twitter").default;
+    case 'down-triangle':
+      return require("./down-triangle").default;
     case 'chevron-right':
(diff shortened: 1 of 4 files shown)


ticky:
🙌

app/components/shared/Icon/index.js
```diff
@@ -35,6 +35,8 @@ const pathNodes = (icon) => {
       return require("./github").default;
     case 'twitter':
       return require("./twitter").default;
+    case 'down-triangle':
+      return require("./down-triangle").default;
```
ticky: huh! this can actually totally be done with webpack wildcards! maybe I’ll hack that in eventually

toolmantim: Oh neat!

keithpitt:
🚀

PR #74, opened 2016-10-05
Fix overflow issue with long build names that don't have spaces

underscorebrody:
BEFORE 
screen shot 2016-10-05 at 7 38 21 am

AFTER
screen shot 2016-10-05 at 7 38 31 am


ticky:
Thanks for your contribution! I had no idea text-overflow worked properly for multi-line elements now!

underscorebrody:
Yeah it's pretty handy! Can't say I tested cross-browser 😅 just Chrome, but it shouldn't get any worse on others at least!

PR #75, opened 2016-10-05
Added support for the new `blocked` state on the Pipelines page

keithpitt:
Blocked builds will now show with the correct icon and will appear in the graph


keithpitt:
OMG YOLO MERGE

PR #76, opened 2016-10-06
Add duration to build info

ticky:
- makes build duration rendering a shared front-end responsibility and consolidates implementations
- adds "micro" duration to build tooltip and my builds dropdown
- switches to snapshot testing
- adds `builds` library


ticky:
ooh, it occurs to me that `builds.js#buildTime` doesn’t explicitly handle the “blocked” state

keithpitt:
Amazing! Only thing we should do is rename `BuildStatus` to `BuildStatusDescription` to just spell out that it's a text based representation of the build.

Once that's changed, 🚀👍

PR #77, opened 2016-10-06
Loading animation for the pipeline graph

keithpitt:
- [x] Finish it off


keithpitt:
Gonna yolo merge this as well into master to see how it feels on production.

PR #79, opened 2016-10-10
Pipelines page — icon/graph tweaks

sj26:
Show scheduled builds as features build, and in the graph:

screenshot 2016-10-10 14 35 26

Also tweaks colours and icons slightly.

/cc @toolmantim

### DIFF
--- a/app/components/organization/Pipeline/graph.js
+++ b/app/components/organization/Pipeline/graph.js
@@ -12,6 +12,8 @@ const RUNNING_COLOR = "#FFBA03";
 const RUNNING_COLOR_HOVER = "#DE8F0C";
 const FAILED_COLOR = "#F83F23";
 const FAILED_COLOR_HOVER = "#AA0A12";
+const SCHEDULED_COLOR = "#aeaeae";
+const SCHEDULED_COLOR_HOVER = "#6e6e6e";
 const PENDING_COLOR = "#DDDDDD";
 const PENDING_COLOR_HOVER = "#DDDDDD";
 
@@ -137,7 +139,9 @@ class Graph extends React.Component {
   }
 
   colorForBuild(build) {
-    if (build.state === "running") {
+    if (build.state === "scheduled") {
+      return SCHEDULED_COLOR;
+    } else if (build.state === "running") {
       return RUNNING_COLOR;
     } else if (build.state === "passed" || build.state == "blocked") {
       return PASSED_COLOR;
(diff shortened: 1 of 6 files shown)


app/components/organization/Pipeline/graph.js
```diff
@@ -191,7 +197,7 @@ export default Relay.createContainer(Graph, {
     pipeline: () => Relay.QL`
       fragment on Pipeline {
         id
-        builds(first: 30, branch: "%default", state: [ BUILD_STATE_RUNNING, BUILD_STATE_PASSED, BUILD_STATE_FAILED, BUILD_STATE_CANCELED, BUILD_STATE_CANCELING, BUILD_STATE_BLOCKED ]) {
+        builds(first: 30, branch: "%default", state: [ BUILD_STATE_SCHEDULED, BUILD_STATE_RUNNING, BUILD_STATE_PASSED, BUILD_STATE_FAILED, BUILD_STATE_CANCELED, BUILD_STATE_CANCELING, BUILD_STATE_BLOCKED ]) {
```
toolmantim: Is this now ALL THE STATES?

keithpitt:
Instead of the 2 different grays, what if they were little empty boxes (if there wasn't a build there?)

app/components/organization/Pipeline/status.js
```diff
@@ -63,7 +63,7 @@ export default Relay.createContainer(Status, {
     pipeline: () => Relay.QL`
       fragment on Pipeline {
         id
-        builds(first: 1, branch: "%default", state: [ BUILD_STATE_RUNNING, BUILD_STATE_PASSED, BUILD_STATE_FAILED, BUILD_STATE_CANCELED, BUILD_STATE_BLOCKED ]) {
+        builds(first: 1, branch: "%default", state: [ BUILD_STATE_SCHEDULED, BUILD_STATE_RUNNING, BUILD_STATE_PASSED, BUILD_STATE_FAILED, BUILD_STATE_CANCELED, BUILD_STATE_BLOCKED ]) {
```
keithpitt: You'll also need to update the Relay preload to include this state.

sj26: Yeah! I can do that when bumping the commit at the other end?

ticky:
Hell yeah, tests!

sj26:
@keithpitt I don't mind! Up to you and @toolmantim.

sj26:
So if pipeline status = last build running then it looks like:

screenshot 2016-10-10 15 32 09

sj26:
And if we do that then we probably want "either the last finished build, or the last build in any state" for a nicer on-boarding state which I can try to figure out in GraphQL.

PR #81, opened 2016-10-10
Differentiate blank and scheduled builds

toolmantim:
Then:

before-1

Now:

after-1


keithpitt:
Nice! I think the labels in the description are around the wrong way, so my brain broke trying to figure out what changed.

ticky:
Fixed the specs for you! :smile:

PR #87, opened 2016-10-11
Fix inconsistent My Builds count on interaction

ticky:
Fixes the issue where the My Builds counter would occasionally show an inconsistent count when you opened the dropdown.

Moves the responsibility for these updates into the Navigation component, where the data is fetched, to avoid weird post-push inconsistency.


app/components/user/BuildsCountBadge.js
```diff
@@ -56,6 +57,12 @@ class BuildsCountBadge extends React.Component {
   };
 
   componentWillReceiveProps = (nextProps) => {
+    // if the next `viewer` instance is the same as our old one, any statistics
+    // attached are *probably* outdated and our previous state is more trustworthy
+    if (nextProps.viewer === this.props.viewer) {
```
toolmantim: @ticky and I just went through this, and this is probably a sign there's some weird smell going on with the modelling here. Because you have to understand the viewer updating/data flow of the Navigation to even touch this file, and this file doesn't do much.

PR #89, opened 2016-10-12
Use yarn instead of npm

ticky:
This’ll not work properly until yarnpkg/yarn#733 is resolved, but yarn seems very good, even this early!


ticky:
4845983 was moved into #97 as it also triggers the eslint update

keithpitt:
I'm with @sj26! Should be fine!

PR #91, opened 2016-10-13
Upload coverage report as artifact

ticky:


keithpitt:
🚀

ticky:
I don’t feel like this is quite finished - having to fetch an arbitrary number of files with no ability to enumerate them makes it kinda useless. I might make it tar them.

PR #98, opened 2016-10-18
Finished -> Passed/Failed

toolmantim:
This changes the tooltip text from:

pasted image at 2016_10_17 12_05 pm

to "Failed Tuesday at 2:33 PM"

Which is important because at the moment, colour is the only way to differentiate the builds in the build graph… there's no icon or words to tell people what the difference between passed and failed graphs are.


ticky:
You’ll need to update some snapshot tests! Let me know if you want help with that :smile:

toolmantim:
@ticky thanks, just spotted that now! Pushed a fix

PR #99, opened 2016-10-20
Pipeline Centric Teams Page

keithpitt:
This PR adds a new page at `/[org]/[pipeline]/settings/teams` that allows the user to manage which teams this particular pipeline is in.

I was able to use all the existing components and mutations to make it all just work (which is rad)

The design of this screen will 100% change when it Teams gets released as a proper feature, but I just wanted to get the structure and functionality in there so it can start to be used by customers.

screen shot 2016-10-20 at 4 12 56 pm


keithpitt:
This page will show you "what teams is this pipeline in" and "what level of access does it have". It also respects team permissions so you won't be able to change stuff unless you're a team admin.

SomeoneWeird:
🎉

PR #101, opened 2016-10-24
Fix graph calculations

ticky:
It turned out the bug in the graph rendering was actually nothing to do with the build _state_ at all, but that newer, quick build graphs were never shrunk by the longer build times following them!

Before:
screen shot 2016-10-24 at 11 00 20 am

After:
screen shot 2016-10-24 at 10 58 20 am

### DIFF
--- a/app/components/organization/Pipeline/graph.js
+++ b/app/components/organization/Pipeline/graph.js
@@ -85,7 +85,13 @@ class Graph extends React.Component {
 
   renderBars() {
     const bars = [];
-    let maximumDuration = 1; // 1 to avoid a `0/0` when we calculate percentages
+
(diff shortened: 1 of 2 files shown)


app/components/organization/Pipeline/graph.js
```diff
@@ -85,7 +85,13 @@ class Graph extends React.Component {
 
   renderBars() {
     const bars = [];
-    let maximumDuration = 1; // 1 to avoid a `0/0` when we calculate percentages
+
+    // `maximumDuration` is wrapped in an object so it's passed by
+    // reference, which means all bars get the final, correct value
+    // despite the generating loop only occuring once
+    const graphProps = {
```
keithpitt: Sneaky ;)

PR #109, opened 2016-11-01
Improve header button wrapping behaviour

ticky:
This allows My Builds to wrap within the menu bar, and remain centered vertically, rather than bleeding off the edge of the menu bar. There's still a lot to optimise in this menu, but this fixes one major bother of mine, at least :P


keithpitt:
Got some before/after screenshots? :)

ticky:
Before:
screen shot 2016-11-01 at 9 55 17 am

After:
screen shot 2016-11-01 at 9 55 35 am

The effect is even more pronounced when there’s a build running:
screen shot 2016-11-01 at 9 57 20 am
screen shot 2016-11-01 at 9 57 27 am

PR #114, opened 2016-11-03
Hide username on mid-sized screens

toolmantim:
Things are getting a little cramped on smaller screens:

boo

This hides the username from medium and small sized screens, and only shows it on extra-small (mobile, when it goes to two rows) and large.

hidden

smaller

### DIFF
--- a/app/components/layout/Navigation/index.js
+++ b/app/components/layout/Navigation/index.js
@@ -119,7 +119,7 @@ class Navigation extends React.Component {
   renderTopOrganizationMenu() {
     if (this.props.organization) {
       return (
-        <span className="flex xs-hide">
+        <span className="flex xs-hide sm-hide md-hide">
           {this.renderOrganizationMenu()}
         </span>
       );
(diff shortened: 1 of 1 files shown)


ticky:
Would putting the “documentation” and “support” items in the dropdown at the medium breakpoint also help?

toolmantim:
Good idea @ticky!

toolmantim:
Ok, I've done another stab at this.

This is the current master:

before

This is the current commit on this PR:

after

The max-width on the names has been made smaller for all resolutions… which means even if there's space, the org or user name might be truncated. But it was the only sane compromise I could do given the class-name style setup we're using.

chrislloyd:
Could you perhaps drop the right arrow next to "Really long…" on smaller screens? It makes less sense when there's nothing next to it.

toolmantim:
@chrislloyd yeah, that's an idea! We haven't dropped the right arrows anywhere yet. I'll keep that in mind next time I take a stab at this.

PR #116, opened 2016-11-04
Use tabular numerals in navbar badges

ticky:
This makes the badges less prone to little changes in layout thanks to changing numbers - they'll only change width if there's a new significant digit (where fitted).

Here’s an image from the CSS fonts level 3 spec, demonstrating proportional and tabular numerals;

example image from spec

And here’s how it changed on Buildkite;

screen shot 2016-11-04 at 2 51 29 pm
🔜
screen shot 2016-11-04 at 2 51 10 pm

### DIFF
--- a/app/components/shared/Badge.js
+++ b/app/components/shared/Badge.js
@@ -8,10 +8,15 @@ class Badge extends React.Component {
   };
 
   render() {
+    const { children, className } = this.props;
+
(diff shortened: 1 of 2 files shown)


keithpitt:
This is a really nice change - I would have never have thought to do this! 🚀

toolmantim:
Nice one! 👍

PR #123, opened 2016-11-11
Support Dialog

keithpitt:
Started playing around with opening a dialog when you hit the "Support" button instead of taking you straight to email.

- [x] How to add images in Frontend?
- [x] Design tweaks
- [x] Fix opacity

nov -15-2016 14-42-31


toolmantim:
AMAZING

keithpitt:
This is ready to go! /cc @ticky @toolmantim you can eyeball the design on my machine if you want.

webpack/config.js
```diff
@@ -170,11 +170,18 @@ module.exports = {
...
       {
-        test: /\.(png|svg|jpg|gif)$/i,
+        test: /\.(svg)$/i,
         loaders: [
           'url-loader?limit=8192',
           'image-webpack?optimizationLevel=7&interlaced=false'
         ]
+      },
+      {
+        test: /\.(jpe?g|png|gif|svg)$/i,
```
ticky: Looks like both rules are applied to svgs!

app/components/layout/Navigation/index.js
```diff
@@ -292,7 +312,9 @@ class Navigation extends React.Component {
 
             {this.renderMyBuilds()}
             <NavigationButton className="py0 xs-hide sm-hide" href={`/docs`}>Documentation</NavigationButton>
-            <NavigationButton className="py0 xs-hide sm-hide" href="mailto:support@buildkite.com">Support</NavigationButton>
+            <NavigationButton className="py0 xs-hide sm-hide" onClick={this.handleSupportClick}>Support</NavigationButton>
```
ticky: This is missing from the dropdown (mobile) version! Speaking of, how does it look on mobile?

webpack/config.js
```diff
@@ -170,11 +170,18 @@ module.exports = {
...
+        loaders: [
+          'file?hash=sha512&digest=hex&name=[hash].[ext]',
+          'image-webpack?bypassOnDebug&optimizationLevel=7&interlaced=false'
+        ]
```
ticky: What’s the rationale for this split? I suspect what we  might _actually_ wanna change is the `url-loader` config, as it passes through to `file-loader` for files larger than `limit`, and I suspect ours is too big!

app/components/layout/Navigation/support-dialog.js
```diff
@@ -0,0 +1,80 @@
...
+  // Using setState in here is a bit yucky, but it's the easiest way to get
+  // ReactCSSTransitionGroup to behave correctly so it shows the animation when
+  // mounting.
+  componentDidMount() {
+    /*eslint-disable react/no-did-mount-set-state */
+    setTimeout(() => {
+      this.setState({ showing: true });
+    }, 0);
+    /*eslint-enable react/no-did-mount-set-state */
+  }
```
ticky: I still don’t think anyone has figured out the “right” way to do this in React, so don’t feel too bad :P

app/components/layout/Navigation/support-dialog.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    return (
+      <div className="background bg-white transition-popup rounded-2 shadow center relative" style={{ padding: "50px 10px", zIndex: 1002, width: 500 }}>
+        <button className="btn absolute circle shadow bg-white bold flex items-center justify-center border border-white p0" style={{ top: -10, right: -10, width: 30, height: 30 }} onClick={this.props.onClose}>
+          <Icon icon="close" title="Close"/>
+        </button>
+
+        <h1 className="bold mt0 mb4"><Emojify text="We’re here to help :wave:"/></h1>
+        <div className="mb4 pt3">{shuffle(PEOPLE).map((person) => this.renderPerson(person))}</div>
```
ticky: Given that the function being called contains no logic it feels like it’d be cleaner just written out in-place with proper newline use!

app/components/layout/Navigation/support-dialog.js
ticky: Since we’re already managing state in the parent component, maybe the transition group should in fact go there? Seems to be the approach used by `BuildsDropdown`

PR #127, opened 2016-11-15
Sneaky local search for agents 🕵🏻‍♀️

ticky:
- when the entire list of agents is loaded, rather than pinging the server for searches, filter the local data set
- filter applies to;
  - agent name
  - agent meta data values (when there’s no `=` in filter)
  - meta data key-values
- also sets the page size to 100 - this means _most_ customers should be exclusively doing local searches most of the time!


toolmantim:
Wee! This will be a muchos better experience for a lot of people.

PR #131, opened 2016-11-25
Clean up the support dialog design

toolmantim:
dialog-after-fo-realz


toolmantim:
I could use the new CSS style component artisanal animation gear!

ticky:
Do it!!! 😃

PR #134, opened 2016-12-05
Make Dropdowns viewport-aware

ticky:
**Before:**
screen shot 2016-12-05 at 12 30 02 pm

**After:**
screen shot 2016-12-07 at 2 05 16 pm

- makes the dropdown viewport-aware
  - Dropdowns with `align="center"` (now the default) will automatically position (and resize, if necessary) themselves to stay within the viewport!
  - they'll make sure their bodies are visible, set the position of their nibs to match their parents, and set their `transformOrigin` to line up (bonus points: `transformOrigin`s for centered Dropdowns were actually previously slightly wrong!)
  - resizing is handled with intelligent debouncing;
    - if a dropdown is shown, it'll update every 100ms until resizing ceases, which seems to be about the right spot for it to feel right
    - if a dropdown isn't shown, we're rather lazier and only update 250ms after the last resize event (this is good because reading DOM layout is expensive!)
- *all* Dropdowns (including those which were previously on-screen already) now use this viewport-aware behaviour
  - we can probably delete the left and right alignment code! ✨
- this introduces a new way of using Styled Components; in an attempt to attain "composition," I've tried out using defaultProps on the component definitions. This means that the styles and the classes they compose are shown in the same location, but it doesn't read quite as well as I'd have hoped - let me know what you think of it!

### DIFF
--- a/app/components/layout/Navigation/index.js
+++ b/app/components/layout/Navigation/index.js
@@ -283,7 +283,7 @@ class Navigation extends React.Component {
               <NewChangelogsBadge className="mr2 relative" style={{ top: -5, marginLeft: -8 }} viewer={this.props.viewer} />
             </span>
 
-            <Dropdown align="left" width={250} className="flex" style={{ minWidth: "5em" }} onToggle={this.handleOrgDropdownToggle}>
+            <Dropdown width={250} className="flex" style={{ minWidth: "5em" }} onToggle={this.handleOrgDropdownToggle}>
               <DropdownButton
                 className={classNames("py0", { "lime": this.state.showingOrgDropdown })}
                 style={{
@@ -309,7 +309,7 @@ class Navigation extends React.Component {
             <NavigationButton className="py0 xs-hide sm-hide" href={`/docs`}>Documentation</NavigationButton>
             <NavigationButton className="py0 xs-hide sm-hide" onClick={this.handleSupportClick}>Support</NavigationButton>
 
-            <Dropdown align="right" width={170} className="flex" ref={(userDropdown) => this.userDropdown = userDropdown} onToggle={this.handleUserDropdownToggle}>
+            <Dropdown width={170} className="flex" ref={(userDropdown) => this.userDropdown = userDropdown} onToggle={this.handleUserDropdownToggle}>
               <DropdownButton className={classNames("py0", { "lime": this.state.showingUserDropdown })}
                 style={{ paddingRight: 0 }}
               >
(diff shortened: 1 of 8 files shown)


keithpitt:
@ticky nice! How does it look in other places the `Dropdown` dooflicky is used? Places I can think of are:

- Organization dropdown
- Teams "change access level" dropdown
- Probably other places you'll need to get `Dropdown` for

ticky:
I’ll get that double-checked!

ticky:
Oops, apparently there’s a bit more to go, there… and it’s a bit complicated

keithpitt:
All good! Wanna throw some TODO's in the Description of what you've got left to do? When there all checked off I know it's review time )

ticky:
It looks like this is going to need my viewport-check technology after all - so I think the plan will be to;

- [ ] Remove this approach, but keep the use of Styled Components
- [ ] On load and on resize of the Dropdown component;
  - [ ] Get the size of the viewport, and the position of the dropdown’s root element
  - [ ] Check if the dropdown would go off-screen, either to the left or right, based on the specified width and the position of the root element
  - [ ] If we’ve not rendered before, or if we’d now be off-screen,
    - [ ] Offset the nib and dropdown appropriately using horizontal margins

ticky:
**New, ✨shiny✨ PR update;**

Now with fancy viewport-aware goodness!

keithpitt:
Nice! Got some screenies? Will look over ze codez now.

ticky:
The screenshots and info in the first post has been updated! :smile:

toolmantim:
Woo, nice work getting this to just use the viewport width!!

Without digging into the code, I've just done some testing. I started by resizing my browser window, I get some offscreen-ness on desktop with the user dropdown:

user dropdown

And for My Builds it becomes slightly off-screen:

scrollbars

I think I see what's happening… There's a funny state situation, where if you reload the page, everything is positioned correctly. But if you resize the page from large to small or small to large, the nibs and positioning are a little off. Is that intended?

One a clean page refresh, everything is positioned perfectly tho 👌🏼

ticky:
Hmm, that _should_ be being handled…

ticky:
Oops! While I was trying to optimise for the other cases I actually broke it, and because hot reload doesn’t reload handlers it didn’t show up for me, haha!

On the topic; should `align="left"`and `align="right"` even continue to be implemented now? I think this functionality plus the `nibOffset` means we should be okay for any future situation.

toolmantim:
Rad! I'll re-check.

No reason to keep the align stuff I think. Let's rm it!

ticky:
`rm`’d! :sparkles:

PR #135, opened 2016-12-06
Implement setup=true behaviour for Agents page

ticky:
Spot the difference;

screen shot 2016-12-05 at 4 15 21 pm
screen shot 2016-12-05 at 4 15 32 pm

👻

### DIFF
--- a/app/components/agent/Index/installation.js
+++ b/app/components/agent/Index/installation.js
@@ -1,16 +1,99 @@
 import React from 'react';
+import Relay from 'react-relay';
+import { seconds } from 'metrick/duration';
+import shallowCompare from 'react-addons-shallow-compare';
 
+import Button from '../../shared/Button';
+import Dialog from '../../shared/Dialog';
+import Emojify from '../../shared/Emojify';
 import Panel from '../../shared/Panel';
 
-export default function AgentInstallation() {
-  return (
-    <Panel className="mb3">
-      <Panel.Header>
-        Installation documentation
-      </Panel.Header>
-      <Panel.Section>
-        See the <a className="blue hover-navy text-decoration-none hover-underline" href="/docs/agent/installation">agent installation documentation</a> for the full details of installing and configuring your Buildkite agents for any machine or architecture.
-      </Panel.Section>
-    </Panel>
-  );
-}
+class AgentInstallation extends React.Component {
+  static propTypes = {
+    organization: React.PropTypes.shape({
+      agents: React.PropTypes.shape({
+        count: React.PropTypes.number.isRequired,
+        edges: React.PropTypes.array.isRequired
+      }).isRequired
+    }).isRequired,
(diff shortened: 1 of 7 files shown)


ticky:
This doesn’t have the `AgentConnectedModal` in it yet!

keithpitt:
@ticky oh right! We have some `Dialog` technology you can use!!

ticky:
## Now with dialogs:

### Old
old and busted

### New
new hotness

keithpitt:
Oh nice! Does this mean it's ready for a code review party?

ticky:
Yep! Go for it 👍

app/components/agent/Index/installation.js
```diff
@@ -0,0 +1,99 @@
...
+    }).isRequired,
+    relay: React.PropTypes.object.isRequired
+  };
+
+  state = {
+    isDialogOpen: false
+  };
+
+  componentDidMount() {
+    this._agentRefreshInterval = setInterval(this.fetchUpdatedData, 5::seconds);
```
keithpitt: Maybe we should use `setTimeout` and then schedule another timeout once that's finished? Something like:

```
function refresh() {
  setTimeout(function() {
    this.props.relay.forceFetch(true, function() {
      this.refresh();
    })
  }, 5::seconds)
}
```

If requests are taking a long time (for some reason) they'll keep piling up, where as with this approach, it'll only schedule the next one after the first one finishes, if that makes sense?

app/components/shared/Dialog.js
```diff
@@ -43,60 +110,71 @@ class Dialog extends React.Component {
...
-        <span className="block fixed flex items-center justify-center" style={{ top: 0, left: 0, bottom: 0, right: 0, zIndex: 1000 }}>
-          <ReactCSSTransitionGroup transitionName="transition-slide-up" transitionEnterTimeout={150} transitionLeaveTimeout={300}>
-            {this.renderDialog()}
-          </ReactCSSTransitionGroup>
-          {this.renderBackdrop()}
-        </span>
-      );
-    } else {
-      return null;
+  maybeClose = (event) => {
```
ticky: `closeMeMaybe`

keithpitt: Probably the best method name ever?

ticky:
We haven’t used `setTimeout` in some other places which poll, so I think that’s a thing we should probably update all at once if we wanna be changing it!

keithpitt:
Meh, it just something I noticed, but in reality I don't think it'd be an issue!

app/components/shared/Dialog.js
ticky:
```javascript
/* heyNowIJustMetYouAndThisIsCrazyButHeresMyCallbackSo*/closeMeMaybe
```

ticky:
It’s definitely a valid criticism, though! I’ll open an issue for it :)

PR #136, opened 2016-12-06
Yarn Upgrade

ticky:
This bumps dependencies and updates usages for some packages with breaking changes! :sparkles:


keithpitt:
Best branch name eva.

PR #137, opened 2016-12-06
Use new "Notice Technology" for the "New Pipelines Page" banner

keithpitt:


app/components/organization/Show.js
```diff
@@ -32,20 +34,24 @@ class OrganizationShow extends React.Component {
...
     return (
       <DocumentTitle title={`${this.props.organization.name}`}>
-        <PageWithContainer>
-          <div className="flex mb2 items-start">
-            <div className="mr-auto flex items-start">
-              <h1 className="h1 p0 m0 mr4 regular line-height-1 inline-block">Pipelines</h1>
-              {this.renderTeams()}
-            </div>
+        <div>
+          <NewPipelinesPageNotice viewer={this.props.viewer} />
```
ticky: Was there a reason this needed to be put outside the `<PageWithContainer />`?

keithpitt: `<DocumentTitle>` (it's old parent) can only have 1 child (boo).

keithpitt: Ohhh, sorry. I get ya now :)

(Trailing on from my rant on this component the other day) The `PageWithContainer` component just does a `<div className="container">` and `.container` restricts the width to the center of the page (960px on some devices). The banner needed to be full width, so I had to put it outside of that component otherwise it's width would have been restricted as well.

If `PageWithContainer` was to just disappear somehow I'd totally :+1: that!

PR #138, opened 2016-12-07
Proof-of-concept: Classed component

ticky:
This was an idea I had while thinking about how to make Styled Components play nice with Basscss, and while it sounds like Styled Components may get support for a class presetting mechanism built-in, this might be a good approach in the mean time.

It allows for supplying the kinds of object interpolations you usually do to ClassNames, but within a callback function you supply. Alternately, you can just supply strings and such.

And as an extra bonus, it allows us to obviate a whole bunch of implementation detail of components which are really basic wrappers around attaching a preset class set over an HTML element.


app/components/shared/PageWithContainer.js
```diff
@@ -1,15 +1,6 @@
...
-    children: React.PropTypes.node.isRequired
-  };
-
-  render() {
-    return (
-      <div className="container">{this.props.children}</div>
-    );
-  }
-}
+const PageWithContainer = classed('div', 'container');
```
keithpitt: This is an interesting idea, and `PageWithContainer` is an interesting use of it as well.

I've actually had the thought lately that some of these "dumb" components we have may be more hassle than they're worth. Let's take this example:

```
render() {
  return (
    <PageWithContainer>
      <div>Content 1</div>
      <div>Content 2</div>
    </PageWithContainer>
  )
}
```

We write this render function looks good. And "oh hey, there's this `PageWithContainer` that makes my life setting up a page with a container easy". I look at it in the browser, and looks good. We've abstracted away what `PageWithContainer` means, so if we want to change it globally, we just change it in one place.

In the future we decide to change `PageWithContainer` to have do `className="container flex"`, all of a sudden, every use of it is now broken. We've kinda broken what it means to be a `PageWithContainer`, so we'd need to make a new one `PageWithContainerFlex` or something..

I'd argue that `PageWithContainer` is actually worse than just putting `<div className="container">` in your render function:

```
render() {
  return (
    <div className="container">
      <div>Content 1</div>
      <div>Content 2</div>
    </div>
  )
}
```

It also means the next time we want to do something custom (like adding `flex`) we don't need to go adding new global components or things like that.

Unless it's a complex set of chained classNames (like Panel) for example, then I don't think it's worth having it abstracted like this! Also having dumb components means when you're onboarded with the codebase, you have to essentially learn an entire visual framework, instead of just using the tools you want how you want to.

And as the person that originally wrote `PageWithContainer` it was from my "Oh I heard about dumb components, it sounds good, we should use them" and not really knowing what a dumb component should be :P And I kinda regret adding it now.

When dealing with visual stuff and flex, you really need to know what the parent element is doing to make everything work. If you're writing a function and you see `PageWithContainer` you'd have to go look up the definition of that class to wonder "hey what does this actually do"...

Anyway, that was more of a rant at this class than at your approach! I think I'm just leaning more towards "dumb components that just have 1-2 class names should just be merged into the original code". I had this same problem when trying to fix something on the agents page. The `Panel` classes weren't working for me, so I had to go read how they work, and essentially copy/paste it's implementation into what I was doing, which kinda defeats the whole thing I think...

ticky: Oh I agree! Some of our use of this is a bit ridiculous, so I think it’s worth having the discussion about replacing it, whether with this or with something simpler, because what I keep finding is that each one of these has been implemented with some small use case in mind, and as soon as you try to use it in a more complex one (i.e. add an extra class to one use of a Panel class), you have to add that functionality yourself, and they all get out of sync!

I suspect the more powerful part of this would be allowing it to compose nicely with Styled Components. But as equivalent functionality seems to be on the way, it may not be so necessary! 😜

ticky:
This probably ain’t going anywhere. Closing to tidy up!

PR #143, opened 2016-12-12
Centralise popover logic

ticky:
I got kinda wheel-spinny on this on Friday so I’m going to put it here so it can either die or get worked on more later.

It’s probably going to be worthwhile to get working for the email popover, but I think the actual functionality for that is more important…

### DIFF
--- a/app/components/organization/Pipeline/bar.js
+++ b/app/components/organization/Pipeline/bar.js
@@ -1,11 +1,13 @@
 import React from 'react';
+import Relay from 'react-relay';
 import shallowCompare from 'react-addons-shallow-compare';
 
 import BuildTooltip from './build-tooltip';
+import AnchoredPopover from '../../shared/Popover/anchored';
 
 import { BAR_HEIGHT_MINIMUM, BAR_WIDTH, BAR_WIDTH_WITH_SEPERATOR, GRAPH_HEIGHT } from './constants';
 
-export default class Bar extends React.Component {
+class Bar extends React.Component {
   static propTypes = {
     href: React.PropTypes.string,
     color: React.PropTypes.string.isRequired,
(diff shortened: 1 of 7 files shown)


app/components/organization/Pipeline/build-tooltip.js
```diff
@@ -35,29 +32,24 @@ class BuildTooltip extends React.Component {
...
-              </Media.Description>
-            </Media>
-          </div>
-        </div>
-      );
-    } else {
-      return <div />;
-    }
+    return (
+      <Media align="top" className="mx2 my1">
```
keithpitt: (...continue from my rant last week, `Media` needs to die)

ticky:
Note that with the latest version of this _all_ Popover-based things are 10 pixels lower down, as they no longer attempt to overlap their parents.

app/components/shared/Popover/index.js
```diff
@@ -0,0 +1,84 @@
...
+        <ScrollZone>
+          {children}
+        </ScrollZone>
+      </div>
+    );
+  }
+}
+
+export AnchoredPopover from './anchored';
+export calculateViewportOffsets from './calculate-viewport-offsets';
```
toolmantim: Woo, I haven't seen this before. This felt weird at first, but I think it's the same as exposing a class method in Ruby… basically, some stateless util function.

It did have me wondering if there was a better API we could expose for `Dropdown` so it didn't need to use it, because it's already using `Popover`?

app/components/organization/Pipeline/bar.js
```diff
@@ -105,3 +112,13 @@ export default class Bar extends React.Component {
     }
   }
 }
+
+export default Relay.createContainer(Bar, {
```
toolmantim: I wonder if this is any problem for performance? I know @keithpitt was scared of so many Relay containers being created, and all the `getFragment` calls? Is there such a thing as "too many Relay containers"? It'd be good to find out…

app/components/shared/Popover/index.js
ticky: Actually, I might’ve missed a trick here… hmm

app/components/organization/Pipeline/bar.js
ticky: There’s only one `getFragment` call for _all_ the `Bar` instances, so that shouldn’t be a problem. As for “too many containers” I guess I can’t say. I suspect there’s no “Relay-friendly” way to make it cleaner, though.

app/components/shared/Popover/index.js
toolmantim: Kewl bananas!

PR #144, opened 2016-12-13
Prompt users to add unknown email addresses on the build page

ticky:
This adds a prompt to let you know that your email addresses aren't configured correctly, and gives you a button for easily adding the email address to your account.

Screenshot of prompt to add an email address


app/components/build/EmailPrompt.js
```diff
@@ -88,15 +88,17 @@ class EmailPrompt extends React.Component {
...
 
   render() {
-    const { email = "error@example.ca" } = this.props.build.createdBy;
+    const { email } = this.props.build.createdBy;
     const notice = this.props.viewer.notice;
 
-    // if the build has no email (this shouldn't happen???)
+    // There won't be an email address if this build was created by a
+    // registered user or if this build just has no owner (perhaps it was
+    // created by Buildkite)
```
ticky: thanks for the clarification ;P

app/components/build/EmailPrompt.js
```diff
@@ -88,15 +88,17 @@ class EmailPrompt extends React.Component {
...
     if (!email) {
       return null;
     }
 
-    // if the user has seen the notice and has been dismissed
+    // If the user has seen the notice and has been dismissed
```
ticky: the nerve of this diff, my oh my

ticky:
This isn’t quite finished but now contains the contents of #143, which could make it tricky to review - diff it locally against `maybe-we-should-centralise-this` for a cleaner diff to peruse.

ticky:
I’ve updated this with a to-do list in the PR message!

ticky:
This no longer has the contents of #143 in it! :sparkles:

app/components/build/UserNameWithEmailPrompt.js
```diff
@@ -0,0 +1,399 @@
...
+    return shallowCompare(this, nextProps, nextState);
+  }
+
+  componentDidMount() {
+    const { createdBy } = this.props.build;
+
+    if (createdBy && createdBy.email && !this.isCurrentUsersValidatedEmail(createdBy.email)) {
+      this.props.relay.setVariables({
+        isTryingToPrompt: true,
+        emailForPrompt: createdBy.email.toLowerCase()
```
toolmantim: Is there any reason we’re doing `toLowerCase` here?

app/components/build/UserNameWithEmailPrompt.js
```diff
@@ -0,0 +1,399 @@
...
+                    <rect x="7" y="9" width="16" height="12" rx="2" />
+                    <polyline points="7 9 15 15 23 9.00146484" />
+                  </g>
+                </svg>
+              </Media.Image>
+              {message}
+            </Media>
+            {buttonContent}
+            <Media className="mx4 my3">
+              <Media.Image className="mr2 center" style={{ width: 32 }} />
```
toolmantim: I'm finding all this `Media` stuff icky? Is there any reason we’re still using it, instead of just using flexbox? This class is getting a bit gnarly with the conditionals, and having variables containing chunks of JSX.

app/components/build/UserNameWithEmailPrompt.js
```diff
@@ -0,0 +1,399 @@
...
+              <Media.Description className="dark-gray mt0"><a className="semi-bold lime hover-lime hover-underline" href="/user/emails">Personal email settings</a></Media.Description>
```
toolmantim: Sorry if I made things worse.

Side note: I wanted to use `<h1>` and `<p>` tags and things, but we don't use them? By nice if we could use semantic tags for accessibility, without having any visual styles applied. Perhaps with a style reset?

app/components/build/UserNameWithEmailPrompt.js
keithpitt: I have an outstanding TODO that's to `rm` the `Media` component. Feel free to just get rid of it! (Most of the dumb components we have are gonna hit the rubbish bin since they end up being messy with conditionals when flex box is so much easier)

toolmantim: I'd be so happy about that… the name `Media` and `Media Image` doesn't mean anything to our business logic. I think it was just a copypasta from the BassCSS examples right?

keithpitt: Yeah do it. Total copy/pasta - back when I was still feeling out how things could/should work.

toolmantim:
Also if the lack of Dismiss buttons feels a bit weird, don't sweat it. I'll check that out tomorrow too.

app/components/build/UserNameWithEmailPrompt.js
ticky: Because the server doesn’t do so when matching prompt dismissals!

ticky: (Because they _could_ be case sensitive, but in this case they are not!)

app/components/build/UserNameWithEmailPrompt.js
ticky: Yeah, I wanna destroy it. `Media.Description` doesn’t actually even do anything but get in the way. It literally contains no classes. It’s just a div.

ticky:
Not having access to the original animations you made I couldn’t really match it up much, but yeah, I learned a lot and it was an interesting process of experimentation!

I’ve added @toolmantim’s to-do list to the first comment so we can track ‘em!

toolmantim:
Thanks @ticky!

toolmantim:
Okay, this is ready to rumble!

The only thing that hasn't been done is to use Pusher or something to detect when they've actually verified the email, and then reload the build page. That would make it feel much more magical.

keithpitt:
That's something we can add pretty easily! If you get it merged into master I can take a look at that bit if you want!

I also just realised that we should probably put this behind a feature flag perhaps to see how it feels on prod?

toolmantim:
Cool, let's get Pusher going then if that's pretty easy 🎉

With the feature flag, that might be tricky given the avatar change? Also, there's no mention of My Builds or anything so we're probably good just to send this live without any other changes :)

app/css/hover-underline.css
```diff
@@ -2,6 +2,11 @@
   text-decoration: underline;
 }
 
+/* nasty terrible specificity hack to make .hover-underline work in old-style pages */
```
keithpitt: I think this bit of a CSS should in the old CSS instead of in frontend?

keithpitt:
I wouldn't want the Pusher stuff hold up this getting into prod! It's easy, but not like 2 lines easy :P (there's some backend changes we'd need to make). It's probably better to just send it live as is (with a feature flag that I've got some notes for in the associated backend PR for this change)

app/components/build/AvatarWithUnknownEmailPrompt.js
```diff
@@ -0,0 +1,337 @@
...
...1953 L16.0205078,17.3251953 L16.0205078,16.7871094 C16.0205078,16.3232399 16.1287424,15.9428726 16.3452148,15.6459961 C16.5616873,15.3491196 16.9729787,15.0244158 17.5791016,14.671875 C18.2099641,14.2945945 18.6954736,13.8508326 19.0356445,13.3405762 C19.3758155,12.8303197 19.5458984,12.2195674 19.5458984,11.5083008 C19.5458984,10.4816029 19.1562539,9.64046552 18.3769531,8.98486328 C17.5976524,8.32926104 16.5740623,8.00146484 15.3061523,8.00146484 C13.9269137,8.00146484 12.8677609,8.35863901 12.1286621,9.07299805 C11.3895634,9.78735709 11.001465,10.6887966 10.9643555,11.7773438 Z M15.0556641,21.7319336 C15.5195336,21.7319336 15.8906236,21.6005059 16.1689453,21.3376465 C16.447267,21.0747871 16.5864258,20.7207054 16.5864258,20.2753906 C16.5864258,19.8300759 16.447267,19.4744479 16.1689453,19.2084961 C15.8906236,18.9425442 15.5195336,18.8095703 15.0556641,18.8095703 C14.5856096,18.8095703 14.209881,18.9425442 13.9284668,19.2084961 C13.6470526,19.4744479 13.5063477,19.8300759 13.5063477,20.2753906 C13.5063477,20.7207054 13.6470526,21.0747871 13.9284668,21.3376465 C14.209881,21.6005059 14.5856096,21.7319336 15.0556641,21.7319336 Z" stroke="none" fill="currentColor" fillRule="evenodd" />
```
keithpitt: Woah. This should probably be in an Icon file!

app/components/build/AvatarWithUnknownEmailPrompt.js
```diff
@@ -0,0 +1,337 @@
...
+  handleVerificationResentSuccess = () => {
+    this.setState({ isSendingVerification: false, hasSentSomething: true });
+  };
+
+  handleLocalDismissClick = () => {
+    this.setState({ hasBeenDismissed: true });
+  }
+
+  getMessages(loading) {
+    const {
```
keithpitt: I actually find declaring all the things like this at the top much harder to read than just calling `this.props.blah` in the function.

Take this line for example:

```
// If we haven't decided to send a query for this yet, don't show anything!
if (!isTryingToPrompt) {
  return {};
}
```

As you scroll your eyes down the function, your constantly having to look back up to the top of the function to see where the value actually came from. Where as:

```
// If we haven't decided to send a query for this yet, don't show anything!
if (!this.state.isTryingToPrompt) {
  return {};
}
```

You brain is instantly: "ah cool, so we're checking state here. rad"

toolmantim: I thought exactly the same thing whilst working on it!

app/css/hover-underline.css
toolmantim: I'm not sure this is even used in this PR?

app/mutations/EmailResendVerification.js
```diff
@@ -0,0 +1,34 @@
+import Relay from 'react-relay';
+
+class EmailResendVerification extends Relay.Mutation {
```
toolmantim: We don't actually use this anymore, so perhaps it should be removed @keithpitt @ticky?

app/components/build/AvatarWithUnknownEmailPrompt.js
toolmantim: Fixed

app/components/build/AvatarWithUnknownEmailPrompt.js
ticky: It’s actually `this.props.relay.variables.isTryingToPrompt`, but I take your point!

app/mutations/EmailResendVerification.js
keithpitt: We'll probably use it as some stage :)

app/components/build/AvatarWithUnknownEmailPrompt.js
```diff
@@ -0,0 +1,321 @@
...
+
+  handleAddEmailClick = () => {
+    this.setState({ isAddingEmail: true });
+
+    const mutation = new EmailCreateMutation({ address: this.props.build.createdBy.email, viewer: this.props.viewer });
+
+    Relay.Store.commitUpdate(mutation, { onSuccess: this.handleEmailAddedSuccess, onFailure: this.handleMutationFailure });
+  };
+
+  handleResendVerificationClick = () => {
```
keithpitt: I do believe this function can be removed!

app/components/build/AvatarWithUnknownEmailPrompt.js
```diff
@@ -0,0 +1,321 @@
...
+
+  handleDismissedSucess = () => {
+    this.setState({ isDismissing: false });
+  };
+
+  handleEmailAddedSuccess = () => {
+    this.setState({ isAddingEmail: false, hasSentSomething: true });
+  };
+
+  handleVerificationResentSuccess = () => {
```
keithpitt: And this - I don't think we do verification sending from the dialog anymore, so it's associated bits and pieces can be ditched!

app/components/build/AvatarWithUnknownEmailPrompt.js
```diff
@@ -0,0 +1,321 @@
...
+        message = (
+          <div>
+            <h1 className="h5 m0 mb1 bold">Email verification needed</h1>
+            <p>Please click the link in the verification email we sent to <strong className="semi-bold">{this.props.build.createdBy.email}</strong>.</p>
+            <p className="dark-gray mt0 dark-gray m0 h7">You can resend the verification email or remove this email address in your <a className="semi-bold lime hover-lime hover-underline" href="/user/emails">Personal Email Settings</a></p>
+          </div>
+        );
+      }
+    } else {
+      // Otherwise, we've got an unknown (to Buildkite) email address on our hands!
```
keithpitt: For some reason I read this and then said this in my head:

app/components/shared/Popover/anchored.js
```diff
@@ -7,16 +7,17 @@ import calculateViewportOffsets from './calculate-viewport-offsets';
 
 export default class AnchoredPopover extends React.Component {
   static propTypes = {
+    alwaysShow: React.PropTypes.bool,
```
keithpitt: I don't think `alwaysShow` is used anywhere anymore?

app/components/shared/Icon/index.js
```diff
@@ -53,37 +53,37 @@ const pathNodes = (icon) => {
...
-        {pathNodes(this.props.icon)}
-      </svg>
-    );
-  }
+  return (
+    <svg viewBox="0 0 20 20" width="20px" height="20px" className={props.className} style={style}>
+      {titleNode(props.title)}
+      {pathNodes(props.icon)}
+    </svg>
+  );
```
ticky: Ooh I didn’t even notice this got converted to stateless! Rad!

PR #145, opened 2017-01-03
Make navigation scrollbars lazy

chrislloyd:
When a use has scrollbars forced on (mouse is plugged in, using a tablet) and `overflow` properties are set to `scroll`, scrollbars are always shown. On OSX by default, most scrollbars sit above the content, whereas when they are shown they collapse the content. With this change, scrollbars are shown when the screen is small and the content is wide, but not triggered in the default case.

**Previously (note the weird gray lines on the secondary navigation):**

screen shot 2017-01-03 at 12 09 22 pm

**After the fix:**

_Wide:_

screen shot 2017-01-03 at 12 04 25 pm

_Thin:_

screen shot 2017-01-03 at 12 04 19 pm

### DIFF
--- a/app/components/layout/Navigation/index.js
+++ b/app/components/layout/Navigation/index.js
@@ -145,7 +145,7 @@ class Navigation extends React.Component {
     if (this.props.organization) {
       return (
         <div className="border-top border-gray lg-hide">
-          <div className="container flex flex-stretch" style={{ height: 45, overflowX: 'scroll' }}>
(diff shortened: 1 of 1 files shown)


keithpitt:
Looks good to me! Easy change - and thanks for the screenies - makes it easy to merge :)

keithpitt:
Shipped!

chrislloyd:
Wow so beautiful now ✨  Thanks!

ticky:
Oops! Good catch!

PR #150, opened 2017-01-19
More explicitly connect git commit email with My Builds

toolmantim:
At the moment if you don't have your emails configured correctly, or you're a new user, this is what you see when you click on My Builds:

my-builds-before

The problem is that "you haven't created any builds yet" could be untrue. They could have created builds, they just haven't added their commit email address to their list of Buildkite email addresses.

my-builds-after

* Handles the case where people aren't using git, but some other SCM and using the API (that's why the `git config user.email` is just an example, and it's referred to as "commit email" instead of "git config email" or "git email".
* Helps to remind them what the command is to find out what git commit email they're using.
* Makes the dropdown a bit wider to fit the text. I wanted to do anyway because it gets a bit cramped and tall sometimes even though there's plenty of real estate. And in case you're wondering, it does  seem to handle mobile/small screens AOK already as well 👌🏼
* Makes the button a primary green one. I'm not 100% on this, but I figured it fits with the push of getting people to link their email up properly to builds, so it's probably better.

### DIFF
--- a/app/components/layout/Navigation/index.js
+++ b/app/components/layout/Navigation/index.js
@@ -173,7 +173,7 @@ class Navigation extends React.Component {
     }
 
     return (
-      <Dropdown width={300} className="flex" onToggle={this.handleBuildsDropdownToggle}>
+      <Dropdown width={320} className="flex" onToggle={this.handleBuildsDropdownToggle}>
         <DropdownButton className={classNames("py0", { "lime": this.state.showingBuildsDropdown })}>
           {'My Builds '}
           <div className="xs-hide">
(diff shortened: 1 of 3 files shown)


toolmantim:
I realised it wasn't very nice for brand new signups who are just clicking around, as opposed to assuming something is wrong because an existing account clicked the menu and wondered where their builds were. So I've updated this with another stab at the copy, and button color, to make it a little less pushy and still work for brand new users "Pfft I haven't done any builds yet. I'm going to just ignore this right now" vs "Hrmm it's asking me a question, that sounds like work…"

after

toolmantim:
Last stab, with a hover colour update too:

one-more-stab

I gave up on rewriting the words to ensure the terminal command snippet didn't wrap.

harrietgrace:
I like the latest words, they're 👌🏻

toolmantim:
Rad. Will ship once I fix `verfified` 🤣

PR #151, opened 2017-01-20
Refactored "My Builds" dropdown Relay Container

keithpitt:
This doesn't change any behaviour of My Builds itself, this just moves the data loading into a Relay Container that's always present on the page. Currently the Relay Container is created/destroyed every time you click the "My Builds" button, which sucks because it makes it hard to be smarter about data loading.

I also moved it from `user/BuildsDropdown` to `layouts/Navigation/MyBuilds` just to be a bit more explicit with what it is, and it's hierarchy.

Making this change opens up a ton of possibilities when it comes to changing how the data is loaded and refreshed on the page.


ticky:
Ooh, good idea!

keithpitt:
@ticky 👍 it also cleans up Navigation a lil-bit and removes a ton of stuff that's only specific to MyBuilds

app/components/layout/Navigation/MyBuilds/index.js
```diff
@@ -0,0 +1,193 @@
...
+      scheduledBuildsCount: payload.scheduledBuildsCount,
+      runningBuildsCount: payload.runningBuildsCount
+    });
+  };
+}
+
+export default Relay.createContainer(
+  CachedStateWrapper(
+    BuildsDropdown,
+    { validLength: 60 * 60 * 1000 /* 1 hour */ }
```
ticky: bonus points: use Metrick to change this to `1::hour` 😜

keithpitt: 💯

app/components/layout/Navigation/MyBuilds/index.js
```diff
@@ -0,0 +1,194 @@
...
+        <Spinner />
+      </div>
+    );
+  }
+
+  handleBuildsDropdownToggle = (visible) => {
+    this.setState({ showingDropdown: visible });
+  };
+
+  handlePusherWebsocketEvent = (payload) => {
```
keithpitt: This code is a little weird, but my next PR will probably change all these bits.

PR #152, opened 2017-01-20
Refactored "My Builds" data loading

keithpitt:
So I've tweaked the data loading of the dropdown to be a bit smarter to avoid the spinner as much as possible.

1. When the page loads, it uses the build numbers from the cache, shows them first.
2. After Pusher has connected, it refetches the latest values from GraphQL.

This ensures that we have the latest values, and we'll be notified when we get new ones.

When it comes to loading the actual builds:

- We start preloading the data on "mouse enter" to reduce the initial spinner time. Doesn't matter if they decide not to load it anyway
- If we receive a "user builds change" Pusher event, we refresh the "builds" and "counts", and render them at the same time. We ignore the data from pusher (which does include the latest build counts), but we'll refetch from GraphQL and use those as the source of truth to keep the UI in sync. So if you're viewing a page, and you see the number change in the nav, you'll know that there's something underneath it for you to look at.

### DIFF
--- a/app/components/layout/Navigation/MyBuilds/index.js
+++ b/app/components/layout/Navigation/MyBuilds/index.js
@@ -15,19 +15,20 @@ import CachedStateWrapper from '../../../../lib/CachedStateWrapper';
 import Build from './build';
 import DropdownButton from './../dropdown-button';
 
-class BuildsDropdown extends React.Component {
+class MyBuilds extends React.Component {
   static propTypes = {
     viewer: React.PropTypes.object,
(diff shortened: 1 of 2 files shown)


keithpitt:
Note that when we receive a "user builds change" pusher event, we'll only refresh the underlying build data if you've open My Builds before. If you've not opened it before, it'll use the regular "mouse enter" logic.

keithpitt:
Cool, yeah works fine locally. Will push to prod to see how it feels

PR #153, opened 2017-01-23
Add missing build state icons

### LINKED ISSUE #147: BuildState icons don't handle cancelling builds
Clicking cancel on a build causes the pipeline page build state icons to disappear for a moment while the build is cancelled. Perhaps canceling should be like pending but red, or something?

toolmantim:
Currently this adds a "Cancelling" build state icon, which is the same as running build but red (the same as cancelled, which is a red cross). So the icon change will be from grey spinning to red spinning.

The current handled states are:

- `pending`
- `scheduled`
- `running`
- `passed`
- `blocked`
- `failed`
- `canceled`
- `canceling`

Are we missing any others?

Closes #147

### DIFF
--- a/app/components/organization/Pipeline/graph.js
+++ b/app/components/organization/Pipeline/graph.js
@@ -20,6 +20,10 @@ const SCHEDULED_COLOR = "#BBB";
 const SCHEDULED_COLOR_HOVER = "#888";
 const PENDING_COLOR = "#DDD";
 const PENDING_COLOR_HOVER = "#DDD";
+const SKIPPED_COLOR = "#83B0E4";
+const SKIPPED_COLOR_HOVER = "#3769A2";
+const NOT_RUN_COLOR = "#83B0E4";
+const NOT_RUN_COLOR_HOVER = "#3769A2";
 
 class Graph extends React.Component {
   static propTypes = {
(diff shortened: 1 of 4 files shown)


keithpitt:
The other 2 build states are `skipped` and `not_run` which we don't have icons for?

toolmantim:
Ok, I've updated this to handle `skipped` and `not_run` too:

not-run-skipped

app/components/organization/Pipeline/graph.js
```diff
@@ -20,6 +20,10 @@ const SCHEDULED_COLOR = "#BBB";
 const SCHEDULED_COLOR_HOVER = "#888";
 const PENDING_COLOR = "#DDD";
 const PENDING_COLOR_HOVER = "#DDD";
+const SKIPPED_COLOR = "#83B0E4";
+const SKIPPED_COLOR_HOVER = "#3769A2";
+const NOT_RUN_COLOR = "#83B0E4";
```
keithpitt: `NOT_RUN_COLOR = SKIPPED_COLOR` makes it a little easier to read I think!

keithpitt:
Aside from my comment, 🚀

toolmantim:
Thanks!

PR #154, opened 2017-01-23
No more Media object

toolmantim:
* Removes the Media object by replacing it with some simple flexbox.
* Fixes avatar height loading in the pipeline hover dialogs (the time stamp could jump down on avatar loading without fixed height/width)
* Remove `nibOffset` now that our dropdown has Automatic Nib-Moving Technology™
* Moves out the duplicated spinny selection options that we use on both the team role chooser and the pipeline permission chooser (see below animation)

jan-23-2017 16-51-09


ticky:
Hmm, I left `nibOffset` intact because that page itself used it to make the menus not shift when used. If you’re cool with the effect that’s fine, though!

PR #156, opened 2017-01-24
Upgrade to Webpack 2

toolmantim:
This upgrades our Webpack setup to v2

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "lint": "eslint .",
     "lint-and-fix": "eslint --fix .",
     "build": "NODE_ENV=development webpack --config webpack/config.js --progress --bail",
-    "build-production": "NODE_ENV=production webpack --config webpack/config.js --progress --bail --optimize-minimize"
+    "build-production": "NODE_ENV=production webpack --config webpack/config.js --progress --bail"
   },
   "devDependencies": {
     "assets-webpack-plugin": "^3.4.0",
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "react-transform-hmr": "^1.0.4",
     "style-loader": "^0.13.1",
     "url-loader": "^0.5.7",
-    "webpack": "^1.14.0",
+    "webpack": "2",
     "webpack-dev-server": "^1.16.1"
   },
   "dependencies": {
--- a/webpack/config.js
+++ b/webpack/config.js
@@ -154,8 +154,8 @@ module.exports = {
       {
         test: /\.css$/i,
         use: [
-          'style-loader',
-          'css-loader',
+          { loader: 'style-loader' },
+          { loader: 'css-loader' },
           {
             loader: 'postcss-loader',
             options: {
@@ -176,13 +176,13 @@ module.exports = {
         test: /\.js$/i,
         exclude: /node_modules/,
         use: [
-          'babel-loader'
+          { loader: 'babel-loader' }
         ]
       },
       {
         test: /\.mdx$/i,
         use: [
-          'babel-loader',
+          { loader: 'babel-loader' },
           { loader: 'markdown-component-loader', options: { passElementProps: true } }
         ]
       },
--- a/webpack/config.js
+++ b/webpack/config.js
@@ -41,7 +41,7 @@ var devTool = IS_PRODUCTION ? "source-map" : "cheap-module-eval-source-map";
 
 var plugins = [
   // Only add the 'whatwg-fetch' plugin if the browser doesn't support it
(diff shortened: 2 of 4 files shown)


ticky:
I have zero idea why this doesn’t build in CI - it works just fine here 😕

ticky:
okay I mean that was a long shot but this makes zero sense

toolmantim:
Ok, this looks ready to go now! The error was because of the double uglify JS plugin

app/components/agent/Index/quick-start.js
```diff
@@ -7,8 +7,6 @@ import classNames from 'classnames';
 import Emojify from '../../shared/Emojify';
 import Panel from '../../shared/Panel';
 
-import 'highlight.js/styles/atom-one-light.css';
```
ticky: Why’d this get moved out of here?

package.json
```diff
@@ -56,7 +56,7 @@
     "react-transform-hmr": "^1.0.4",
     "style-loader": "^0.13.1",
     "url-loader": "^0.5.7",
-    "webpack": "^1.14.0",
+    "webpack": "2",
```
ticky: We should probably pick a particular version

app/components/agent/Index/quick-start.js
toolmantim: Webpack couldn't compile it, and it's the only file that requires CSS outside of `main.css` so figured I'd move it in there. But I'm not sure where we're heading with stuff like this? 

I was going to do a follow-up PR w/ `ExtractTextPlugin` and a proper external CSS file anyways, rather than the JS-loaded version we use now. And so this might possibly work again in that setup.

package.json
toolmantim: Pre-yarn I would have agreed, but because versions are locked in `yarn.lock` and yarn will never downgrade a package unless told so, this is probably the most readable/simplest setup? 2 means any "2.x" version which sounds good to me.

We could pick `^2.0.0` which means "any 2.x.x version greater than or equal to 2.0.0" and have a different version in yarn.lock as well… but it'd be exactly the same no?

app/components/agent/Index/quick-start.js
toolmantim: Seems to work putting it back! I'll do that and rebase.

ticky: I prefer having it here so it’s very obvious it’s _only_ used here. We can’t prefix it with anything much as I’d like to, so if it’s at least only loaded when this component is, that’s for the best, IMO!

package.json
ticky: You’ve got a good point! Any idea what Yarn’ll do when you `yarn upgrade` it?

toolmantim: Yeah, I watched with curiosity when I last `yarn upgrade`'d… it just left the `package.json` alone and updated the `yarn.lock` file.

ticky: Ah cool, I guess it’d only write it if the requested version no longer matched the one in `Yarn.lock`. Awesome! 😃

toolmantim: 🤠

toolmantim:
Should be ready to go!

ticky:
(rebased against master so it can be reviewed against current backend)

webpack/config.js
```diff
@@ -154,48 +151,64 @@ module.exports = {
...
+            loader: 'postcss-loader',
+            options: {
+              ident: 'postcss',
+              plugins: function() {
+                return [
+                  require("postcss-import")(),
+                  require("postcss-cssnext")({ features: { rem: false } }),
+                  require('postcss-easings')(),
+                  require("postcss-browser-reporter")(),
+                  require("postcss-reporter")()
```
ticky: clearly needs some [`postcss-stronk`](https://github.com/ticky/postcss-stronk)

webpack/config.js
```diff
@@ -154,48 +151,64 @@ module.exports = {
...
+        use: [
+          { loader: 'babel-loader' }
+        ]
       },
       {
         test: /\.mdx$/i,
-        loader: 'babel-loader!markdown-component-loader?passElementProps=true'
+        use: [
+          { loader: 'babel-loader' },
+          { loader: 'markdown-component-loader', options: { passElementProps: true } }
```
ticky: this is so much nicer, I’m glad Webpack 2 is moving away from the query string nonsense.

webpack/config.js
toolmantim: Hah! 💪🏼

webpack/config.js
toolmantim: Isn't it!

PR #157, opened 2017-01-26
Fix AgentsCount inconsistencies

ticky:
This updates the AgentsCount component to be Relay-aware, and instead of alternatingly trusting the Pusher agent count and Relay agent count, instead takes a change on the Pusher end as a cue to `forceFetch` updated values from Relay.

This causes the component to stay pretty stable when it comes to repaints.

I also took the opportunity to switch over to using `shallowCompare` in `shouldComponentUpdate` and remove the component from the Webpack bridge in app.js

Fixes https://3.basecamp.com/3453178/buckets/1445180/todos/323608085

### DIFF
--- a/app/components/organization/AgentsCount.js
+++ b/app/components/organization/AgentsCount.js
@@ -1,4 +1,6 @@
 import React from 'react';
+import Relay from 'react-relay';
+import shallowCompare from 'react-addons-shallow-compare';
 
 import PusherStore from '../../stores/PusherStore';
 
@@ -7,11 +9,11 @@ import { formatNumber } from '../../lib/number';
 class AgentsCount extends React.Component {
   static propTypes = {
     organization: React.PropTypes.shape({
-      id: React.PropTypes.string.isRequired,
       agents: React.PropTypes.shape({
         count: React.PropTypes.number.isRequired
       })
-    })
+    }),
+    relay: React.PropTypes.object.isRequired
   };
 
   state = {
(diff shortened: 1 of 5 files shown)


app/components/organization/AgentsCount.js
```diff
@@ -7,52 +9,61 @@ import { formatNumber } from '../../lib/number';
...
+    PusherStore.off("organization_stats:change", this.handlePusherWebsocketEvent);
   }
 
-  handleStoreChange = (payload) => {
-    this.setState({ agentCount: payload.agentsConnectedCount });
+  // Like in MyBuilds, we don't take the Pusher data for granted - we instead
+  // take it as a cue to update the data store backing the component.
+  handlePusherWebsocketEvent = (payload) => {
+    if (this.state.agentCount !== payload.agentsConnectedCount) {
+      this.props.relay.forceFetch();
```
keithpitt: Rad - I think we may want to throttle this particular `forceFetch`. If a customer connects 1000 agents, and they all come on one by one, the browser will do 1000 odd requests.

Perhaps a throttling mechanism like:

1. `forceFetch` is called
2. Can't be called again for another 3 seconds
3. If a `forceFetch` is requested during the timeout period, one will be scheduled in when the timer runs out
4. Go to 1

Previously I would have just used the methods from `underscore`, but I don't really wanna introduce that package into frontend. Maybe there's a generic throttling lib we can use? I've wanted to throttle stuff in other places as well.

keithpitt: Just as a follow up, using that mechanism will mean the agents count is only ever updated ever 3 seconds. So if 1000 agents come online, and it takes them 20 seconds, at most well hit the backend 6 or so times.

ticky: [This](https://github.com/component/throttle/blob/master/index.js) looks pretty comprehensive, and pretty tiny!

Also lol @ the size of [individual lodash packages](https://github.com/lodash/lodash/blob/4.1.1-npm-packages/lodash.throttle/index.js)

keithpitt: @ticky `component/throttle` looks perfect! "Supports leading and trailing invocation." is awesome as well.

keithpitt:
Does this also fix the "click the agent button and get a cached version" bug?

ticky:
Yes! The reason that was happening was that when we clicked we would switch back to the version of the data the organisation graph node has, which could’ve been outdated.

app/components/organization/AgentsCount.js
```diff
@@ -1,58 +1,80 @@
...
 
-  handleStoreChange = (payload) => {
-    this.setState({ agentCount: payload.agentsConnectedCount });
+  // Like in MyBuilds, we don't take the Pusher data for granted - we instead
+  // take it as a cue to update the data store backing the component.
+  handlePusherWebsocketEvent = ({ agentsConnectedCount }) => {
+    // We need a "global" last agents connected count so we only ask it to update
+    // once per changed count. This prevents calls from multiple AgentsCount
+    // components triggering repeated forceFetch calls for one Pusher event
+    if (AgentsCount.lastAgentsConnectedCount !== agentsConnectedCount) {
```
keithpitt: Sneaky ;)

keithpitt: But perfect.

ticky: feels like the cleanest way to do this without making it Relay’s responsibility 😅

Issue #139, opened 2016-12-07
Polling components use `setInterval` rather than `setTimeout`


ticky:
On poorly-performing networks this could cause issues!

- [x] `app/components/agent/Show.js` (#235)
- [x] `app/components/agent/Index/installation.js` (#235)
- [x] `app/components/agent/Index/agents.js` (#158)

Noted by @keithpitt in #135

PR #158, opened 2017-01-27
Use push events for Agents List

ticky:
This switches the Agents list to update when it gets a push event. It also adjusts the automatic refresh interval to 15 seconds, and will only fetch once per 5-second window.

Fixes https://3.basecamp.com/3453178/buckets/1445180/todos/353039886, and partially addresses #139


keithpitt:
So just looking at the code, is that "it'll update every 15 seconds, or when we receive a "pusher change count" - which ever comes first?"

ticky:
That’s right, yeah!

PR #159, opened 2017-01-27
Update Agent list relay preload data

ticky:
Partially addresses https://3.basecamp.com/3453178/buckets/1445180/todos/323608010


keithpitt:
Oh! Was the page blocking loading that extra bit of data? What even is that?

ticky:
It’s the data used for the “You just connected an agent!” pop-up. This removes one query, though there are still unfortunately several going on on the page.

keithpitt:
Oh right! Should we be loading that if the agent page isn't in `?setup=true` mode?

ticky:
Hmmm, good point… I’ll see if I can make that happen instead…

PR #161, opened 2017-01-30
Only show the email prompt if theres no associated user

keithpitt:
Before we'd show the prompt even if there was an associated user. This changes the code so it'll only prompt if GraphQL says: "hey, we actually don't have a user here"


ticky:
Wait, why did the registered user fragment get the email added?

keithpitt:
@ticky I dunno! https://github.com/buildkite/frontend/commit/42d93c1b722c35b85bfc7b4bde6b8520ab8fb019 :P

ticky:
Welllll then

PR #163, opened 2017-01-31
Make the Changelogs badge bulletproof

ticky:
Previously, it suffered from the same state caching issues as other Pusher-powered things that could also be interacted with (opening, for instance, the Organizations dropdown, would cause it to use the cached state), this updates it to act correctly in all cases.

It also makes it a relay container which accepts the isMounted variable from its parent, so that it can force an update of only its data set whilst also being updated on-mount without triggering a separate request for its data!

### DIFF
--- a/app/components/layout/Navigation/index.js
+++ b/app/components/layout/Navigation/index.js
@@ -24,7 +24,6 @@ class Navigation extends React.Component {
     relay: React.PropTypes.object.isRequired
   };
 
-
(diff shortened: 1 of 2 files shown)


keithpitt:
I think the footer has a changelog number as well?

ticky:
It does, but it’s managed separately, and doesn’t actually hook up to Pusher!

PR #164, opened 2017-01-31
Some random avatar prompt code cleanups

keithpitt:
Just looking into adding some Pusher gear and noticed some bits and pieces that could be cleaned up  + commented a little.

### DIFF
--- a/app/components/build/AvatarWithUnknownEmailPrompt.js
+++ b/app/components/build/AvatarWithUnknownEmailPrompt.js
@@ -46,7 +46,9 @@ class AvatarWithUnknownEmailPrompt extends React.Component {
     hasBeenDismissed: false
   };
 
-  getUserEmail(email) {
+  // Given an email address, try and return the associated UserEmail GraphQL
+  // node
+  findUserEmailNode(email) {
     const userEmails = this.props.viewer.emails.edges;
 
     const userEmail = userEmails.find(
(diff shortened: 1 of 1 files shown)


app/components/build/AvatarWithUnknownEmailPrompt.js
```diff
@@ -46,7 +46,9 @@ class AvatarWithUnknownEmailPrompt extends React.Component {
     hasBeenDismissed: false
   };
 
-  getUserEmail(email) {
+  // Given an email address, try and return the associated UserEmail GraphQL
+  // node
+  findUserEmailNode(email) {
```
keithpitt: `getUserEmail` reminded me more of a "getter" type method - so I switched it to `find` to kind of hint that you may get nil, and to be more indicative of what it's doing.

app/components/build/AvatarWithUnknownEmailPrompt.js
```diff
@@ -60,19 +62,36 @@ class AvatarWithUnknownEmailPrompt extends React.Component {
...
   isCurrentUsersValidatedEmail(email) {
-    const userEmail = this.getUserEmail(email);
+    const userEmail = this.findUserEmailNode(email);
+
     return userEmail && userEmail.verified;
   }
 
+  // Determines if this paticular build should prompt the current user to add
+  // it's associated email to their user account
+  isUnregisteredWithEmail(createdBy) {
```
keithpitt: I just extracted this logic out to a method just to clean up the `componentDidMount` gear a little bit

app/components/build/AvatarWithUnknownEmailPrompt.js
```diff
@@ -94,13 +113,6 @@ class AvatarWithUnknownEmailPrompt extends React.Component {
   }
 
   getMessages(loading) {
-    // There won't be an email address if this build was created by a
```
keithpitt: I believe this code here was for when we didn't request the email for the `User` type. I believe this logic was to prevent the dialog for not showing once it's been linked to a user. Since we now do the `UnregisteredUser` check, I believe we can just blow this away!

ticky: Yeah, you’re right! Sorry I lost track of that!

app/components/build/AvatarWithUnknownEmailPrompt.js
ticky: 👍🏼

PR #169, opened 2017-02-03
Tweaked data loading for Agents Show

keithpitt:
Tweaked loading of the Agents Show page so have the browser just hang on you while it's performing necessity queries to load the page.

### DIFF
--- a/app/components/agent/Show.js
+++ b/app/components/agent/Show.js
@@ -52,9 +52,7 @@ class AgentShow extends React.Component {
   }
 
   componentWillUnmount() {
-    if(this._agentRefreshInterval) {
-      clearInterval(this._agentRefreshInterval);
-    }
+    clearInterval(this._agentRefreshInterval);
   }
 
   fetchUpdatedData = () => {
--- a/app/components/agent/Show.js
+++ b/app/components/agent/Show.js
@@ -19,7 +19,7 @@ class AgentShow extends React.Component {
   static propTypes = {
     agent: React.PropTypes.shape({
       id: React.PropTypes.string,
-      name: React.PropTypes.string.isRequired,
+      name: React.PropTypes.string,
       connectionState: React.PropTypes.string,
       permissions: React.PropTypes.shape({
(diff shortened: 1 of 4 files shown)


ticky:
I think my main question with this is, what happens if Shopify load this page? Are we then requesting _all_ the agent data repeatedly?

ticky:
wait, never mind, I made assumptions about your approach ;P

app/components/agent/Show.js
```diff
@@ -34,11 +35,22 @@ class AgentShow extends React.Component {
   };
 
   componentDidMount() {
-    this._agentRefreshInterval = setInterval(this.fetchUpdatedData, 5::seconds);
+    // Only bother setting up the delayed load and refresher if we've got an
+    // actual agent to play with.
+    if(this.props.agent && this.props.agent.id) {
+      this._agentRefreshInterval = setInterval(this.fetchUpdatedData, 5::seconds);
```
ticky: wanna update this to do `setTimeouts` instead as part of this?

app/components/agent/Show.js
```diff
@@ -34,11 +35,22 @@ class AgentShow extends React.Component {
...
+      // to show this page.
+      this.props.relay.setVariables({ isMounted: true });
+    }
   }
 
   componentWillUnmount() {
-    clearInterval(this._agentRefreshInterval);
+    if(this._agentRefreshInterval) {
+      clearInterval(this._agentRefreshInterval);
+    }
```
ticky: `clearInterval` doesn’t actually care if it’s given nothing! which is weird, I guess

app/components/agent/Show.js
keithpitt: I was gonna, but then `this.props.relay.setVariables({ isMounted: true })` needs to be in the `if` as well, so I figured this'd be the best place.

app/components/agent/Show.js
keithpitt: Oh lol! Will remove the check.

PR #170, opened 2017-02-06
Make it clear when assuming a user

toolmantim:
This adds an `assuming` prop to `Navigation`, so we can make it clearer when you're assuming a user:

assuming-user

### DIFF
--- a/app/components/layout/Navigation/index.js
+++ b/app/components/layout/Navigation/index.js
@@ -32,7 +32,7 @@ class Navigation extends React.Component {
     showingOrgDropdown: false,
     showingUserDropdown: false,
     showingSupportDialog: false,
-    assumed: window['_adminIsAssumingUser']
+    warning: window['_navigation'] && window['_navigation']['warning']
   };
 
   handleOrgDropdownToggle = (visible) => {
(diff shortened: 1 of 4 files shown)


app/css/background-colors.css
```diff
@@ -19,3 +19,13 @@
 .bg-fuchsia { background-color: var(--fuchsia) }
 .bg-purple  { background-color: var(--purple) }
 .bg-maroon  { background-color: var(--maroon) }
+
+.bg-warning-stripes {
```
toolmantim: I'm not stoked with putting this here…

app/components/layout/Navigation/index.js
```diff
@@ -21,7 +21,8 @@ class Navigation extends React.Component {
   static propTypes = {
     organization: React.PropTypes.object,
     viewer: React.PropTypes.object,
-    relay: React.PropTypes.object.isRequired
+    relay: React.PropTypes.object.isRequired,
+    assumed: React.PropTypes.bool
```
toolmantim: Not sure how to pass this in from where the `Relay.RootContainer` is instantiated

keithpitt: I'd probably do something super basic like: `window._navigation = { warning: true }`. Check for that instead of `this.props.assumed`, and that'd be super simple to implement in the backend.

app/css/background-colors.css
ticky: Easy hack, rename it to `backgrounds.css` lol

toolmantim: Hah! Actually I think that's kinda the solution. Just pushed a new `backgrounds.css`

app/components/layout/Navigation/index.js
toolmantim: I just added something super basic like you said.

I wanted to add it to `Viewer` but I couldn't get access to the session from `Tenant` :|

toolmantim:
@keithpitt I've updated this to detect the global instead

PR #172, opened 2017-02-12
Fix awkward timestamp line breaks in build popovers

toolmantim:
For older builds with larger timestamps, the pipelines build popover width can be too narrow to fit the last line:
feb-12-2017 13-03-07

Instead of using the default `AnchoredPopover` width of 250, this sets it explicitly to 300 which covers most cases for the last line.


keithpitt:
Oh, actually looking at this again, I don't think this includes the popups for the ones that come from the bars. I think this is just for the one that the health icon on the left has. You probably need to change here as well: https://github.com/buildkite/frontend/blob/master/app/components/organization/Pipeline/bar.js#L72

toolmantim:
Nice, good catch. Thanks @keithpitt!

PR #173, opened 2017-02-13
Made pipeline tooltips 300px wide too

toolmantim:
Missed in #172, this makes sure the pipeline graph build tooltips are also wide enough to fit timestamps

### DIFF
--- a/app/components/organization/Pipeline/bar.js
+++ b/app/components/organization/Pipeline/bar.js
@@ -69,7 +69,11 @@ class Bar extends React.Component {
 
     if (this.props.href) {
       return (
-        <AnchoredPopover position="absolute" style={{ height: '100%', left: this.props.left, bottom: 0 }}>
+        <AnchoredPopover
+          position="absolute"
+          style={{ height: '100%', left: this.props.left, bottom: 0 }}
+          width={300}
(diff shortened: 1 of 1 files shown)


app/components/organization/Pipeline/bar.js
```diff
@@ -69,7 +69,11 @@ class Bar extends React.Component {
 
     if (this.props.href) {
       return (
-        <AnchoredPopover position="absolute" style={{ height: '100%', left: this.props.left, bottom: 0 }}>
+        <AnchoredPopover
+          position="absolute"
+          style={{ height: '100%', left: this.props.left, bottom: 0 }}
+          width={300}
```
ticky: given that both places which use `AnchoredPopover` now have a default of 300, maybe we should update `AnchoredPopover`’s defaults instead?

toolmantim: Wise. Will change!

toolmantim: Actually, it should really be hard coded in these two places, because it's very much dependent on the size of the content inside the AnchoredPopover. Perhaps the AnchoredPopover shouldn't have a default size at all?

PR #175, opened 2017-02-22
Clean up agent setup

toolmantim:
- [x] Review + fix agent index
- [x] Review + fix guides
- [x] Check screen sizes

Agent setup before/after:

before-2

before-2-after

AWS tab before/after:

before-1

before-1-after

Agent show page before/after (the after doesn't have the job warning message, because now that is only shown if the agent actually has a job assigned):

agent-before

agent-after-2

### DIFF
--- a/app/components/agent/Index/agent-tokens.js
+++ b/app/components/agent/Index/agent-tokens.js
@@ -64,7 +64,7 @@ class AgentTokens extends React.Component {
   }
 
   renderDescription(token) {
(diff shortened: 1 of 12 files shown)


ticky:
Ooh, love the agent show page change! Much nicer 😃

> Only render descriptions if there's more than 1 token

I’m pretty sure I had that to begin with? When did that go?

toolmantim:
> I’m pretty sure I had that to begin with? When did that go?

I think @keithpitt admitted guilt :)

PR #179, opened 2017-02-22
Improved agent searching

toolmantim:
Changes to remote search (> 100 agents):

* Sends the query to the server on every keyup
* Only shows a spinner if the search is taking > 1 second
* The spinner now lives next to the search field
* Fixes a bug where local search could be triggered by accident

Changes to local search (< 100 agents):

* Filters immediately on keyup
* Fixes the matching logic to match the backend/remote matching logic

### DIFF
--- a/app/components/agent/Index/agents.js
+++ b/app/components/agent/Index/agents.js
@@ -128,30 +128,67 @@ class Agents extends React.Component {
     const { organization: { agents: { edges: agents } = { } } } = this.props;
     const { localSearchQuery } = this.state;
 
-    // return all agents if the search query is falsy
     if (!localSearchQuery) {
       return agents;
     }
 
-    // otherwise, filter by name and metaData
+    // Parse the search. We only want to do this once, outside of the agent
+    // loop, in case they have lots of agents
+    const queries = localSearchQuery.map((string) => {
+      return {
+        string: string.toLowerCase(),
+        metaDataKey: string.split('=')[0],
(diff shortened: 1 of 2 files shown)


toolmantim:
@keithpitt nice, thanks for the hints! I've been able to get it working:

search

It'd be nice to only show the spinner if the remote search has been happening for more than 750ms+, but I couldn't quite figure out how to wrangle that.

ticky:
I feel like `debounce` would still be the better fit here; it’s much less likely the user wants the results after their _first_ input. The 500ms might just be too long. Totally with you on changing the loading spinner behaviour!

toolmantim:
Yeah, the `first` keyup is probably unnecessary, but the rest of the interaction (backspacing, correcting typos, seeing what `osx` vs `osx=true` does, etc) feels snappier w/ throttle.

Unless you're imagining debounce is short enough that it happens between a normal human's keystrokes, like the Safari autocomplete seems to be.

toolmantim:
This is ready, and I've updated the description with what it changes.

For the lack of debounce/throttling, as long as relay cancels the previous `forceFetch` request then it shouldn't make a different to the UX by sending more frequent remote calls (except that there's less delay to wait for the list to filter).

I'd be keen to give this a test on prod with a large number of agents and a realistic latency though.

keithpitt:
Cool, let's get it on pod!

keithpitt:
*prod even

PR #183, opened 2017-02-23
🎊 New "First Agent Connected" dialog

toolmantim:
first-agent-confetti

(padding was slightly tweaked after the video was taken)


toolmantim:
I'm not happy about the `window.location` hack but couldn't think of a better solution. It started because when you hit "(X)" you couldn't see the agent that connected. And also if refresh the setup page with an agent connected, they still can't see their agent.

One idea to get rid of the `window.location` hack would be if we added the agent list to the bottom of the setup page.

PR #188, opened 2017-02-27
Sticky team selections: Follow up, Handle GraphQL failures from missing teams

ticky:
- [x] Handle failures and redirect to “All teams”
- [x] Flash the user a message when redirecting

### DIFF
--- a/app/components/organization/Pipelines.js
+++ b/app/components/organization/Pipelines.js
@@ -56,6 +56,10 @@ class OrganizationPipelines extends React.Component {
           // updated by the parent setVariables call!
           this.props.relay.setVariables({ isMounted: true, teamSearch: null }, (readyState) => {
             // flash error once we've got data so it behaves more like its backend counterpart!
+            //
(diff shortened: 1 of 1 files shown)


app/components/organization/Pipelines.js
```diff
@@ -35,13 +39,22 @@ class OrganizationPipelines extends React.Component {
...
+      if (done) {
         // Now kick off a full reload, which will grab the pipelines again, but
         // this time with all the graph data.
         setTimeout(() => {
           this.props.relay.forceFetch({ includeGraphData: true });
         }, 0);
+      } else if (error) {
+        // if we couldn't find that team in GraphQL, let's redirect to not requesting a team!
+        if (error.source.errors.some(({ message }) => message === 'No team found')) {
+          this.context.router.push(`/${this.props.organization.slug}`);
```
keithpitt: I think we should hit up the `FlashesStore` thingy here with a nice message as to what just happened.

ticky: Ah, cool! I wasn’t sure how to address that! 😃

PR #201, opened 2017-03-16
Remove `react-transform` and friends

keithpitt:
We're using them for hot reloading but those projects have been deprecated and don't work with Webpack 2


keithpitt:
Next step will be to add in this in: https://github.com/gaearon/react-hot-loader

sj26:
ohmy~

PR #208, opened 2017-03-17
Added react-hot-loader

keithpitt:
- [ ] Add react-hot-loader
- [ ] Try and get error handling working


keithpitt:
This works with simple updates, but if you add an error to a `render` function - everything blows up. Closing this and using old tools for now.

PR #212, opened 2017-03-21
A basic YAML editor form component

keithpitt:
I didn't bother looking into delaying the loading of the component. I'll wait to find out what the actual impact on the vendor lib is before doing that. After gzipping it may be only 20kb or something super small.


keithpitt:
Adding `codemirror` only adds an extra 50kb to the vendor package. I'm OK with that!

PR #216, opened 2017-03-23
Add @lox to the support dialog

keithpitt:
screen shot 2017-03-23 at 4 44 44 pm


ticky:
IMO this UI looks better with uneven numbers of team members so we should be sure to take that into account 😜

PR #220, opened 2017-03-29
Let flexbox manage the navbar

ticky:


keithpitt:
@ticky tell us more about your hypotheses with this experiment! As the words you use sound exciting :)

ticky:
We want a few things from the nav bar;
1. For the primary organization navigation to wrap to a second line on small screens
2. For user and organization names to truncate if things get too crowded

I *think* we can just have Flexbox do these things (meaning one copy of the organization nav), for the most part. In its current state, this PR should non-destructively have 2 managed entirely by flexbox, and 1 is currently not yet done.

ticky:
I’ve updated this. It still only does #2, but I’ve checked it’s all sensible and it seems pretty solid!

toolmantim:
I'm just doing some checks of this for you now, sorry I didn't do it yesterday!

toolmantim:
I've just going to paste the layout bugs in here if/as I find them.

Here's one…

long-names

toolmantim:
On other pages, its busts out to the side…

other-pages

toolmantim:
This isn't a bug, but something I just noticed.

Originally, the arrow on the right side of the organization dropdown was kind of pointing to Pipelines/Agents/Users/Settings for the org. But when we switch to two rows on smaller screens, it doesn't really make sense.

org-selector-arrow

Is it possible to hide the right arrow border when we switch to the small screen layout?

ticky:
Good catches! I’ll have a look at fixing them up! 😃

ticky:
I believe we can hide that arrow if we wish, yeah!

toolmantim:
That's all I could find. The rest of it looks ace bananas! 👍🏼

toolmantim:
Sorry, but I found a few more weird things…

very-long-orgs

very-long-orgs-2

very-long-orgs-3

ticky:
Wtf, somehow I’ve lost the change which removed the arrow from this… oh dear.

ticky:
> fixed a case where the "My Builds" button stretched and split over two lines

I actually put that back because otherwise, on very small phones, you lose a bunch of space in the organization dropdown;

screen shot 2017-06-05 at 10 16 09 am screen shot 2017-06-05 at 10 16 24 am

ticky:
That said, I think I’m happy enough for this to go out for now!

toolmantim:
Oh interesting! I was getting it wrap when the org name or user name was large, on the single column version on wider screens. Hrmm…

PR #222, opened 2017-03-30
Pagination of Team Pipelines & Members

ticky:
This adds pagination and search to the team member and pipeline lists.


ticky:
This is ready to push live for customers and some final optimisations based on observations there, I think!

PR #224, opened 2017-04-03
Remove react-type-snob from the `vendor` bundle

ticky:
This prevents an extraneous error from being thrown in production. `react-type-snob` is neither referred to nor used in production and can thus be safely removed!

Also adds a throw when no `EMOJI_HOST` is set, which means you get a failure after 0.5 seconds rather than 90.


ticky:
Possibly! But it’s all okay now :smile:

PR #230, opened 2017-04-07
SSO Tab

keithpitt:
- [x] Show if SSO is enabled
- [x] Add behind feature flag (teams one maybe?)
- [x] Implement new design
- [x] Update with copy tweaks for production use


keithpitt:
🚀

PR #231, opened 2017-04-10
[Experiment] Update icon component to use raw SVG data

ticky:
It’s bugged me for a while that this is using React and not just SVG text. Turns out the gain in the bundle is uh, not that much (28.94 KB vs 20.41 KB) but on the plus side this means we’re relying on Webpack rather than arbitrary lists in the component file to figure out what’s available!


toolmantim:
I'm cool with this change! How does the raw-loader work in prod?

ticky:
Things that are raw-loaded are included in the bundle as usual. The difference being they’re SVG text rather than JS. We could take this further and use something like `svgo` to save a few more bytes but I suspect that wouldn’t be too important (it’d just be trimming whitespace at this point).

toolmantim:
Sweet, let's do it!

PR #232, opened 2017-04-11
Clean up type sizes

toolmantim:
As part of working on the new pipeline header updates, I updated these font-sizes. I'm breaking these sorts of changes out into their own PRs so we can ship them first.

This one updates our type scale to be more sane. `.h4` is now "regular sized text", and we actually define a `.h2` size.

Whilst working on the sizes I needed a style guide I could change and look at, and I left it in there with the idea that we'd mount it to `_styleguide` in development. I just gave up on making that actually work.


ticky:
yesssss!

ticky:
Oh I was just looking at this :P

PR #234, opened 2017-04-11
Clean up "New Build" dialog

toolmantim:
Before:

old

After:

new

(the autofocus still works… I just expanded the Options before hitting the screenshot, and lost focus on the commit message field)

### DIFF
--- a/app/components/pipeline/CreateBuildDialog.js
+++ b/app/components/pipeline/CreateBuildDialog.js
@@ -15,7 +15,7 @@ class CreateBuildDialog extends React.PureComponent {
   };
 
   state = {
-    showingMoreOptions: false,
+    showingOptions: false,
     creatingBuild: false
   };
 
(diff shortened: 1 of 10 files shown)


app/components/pipeline/CreateBuildDialog.js
```diff
@@ -77,50 +75,29 @@ class CreateBuildDialog extends React.PureComponent {
...
+            <CollapsableArea label="Options" maxHeight={250}>
+              <FormTextarea
+                name="build[env]"
+                label="Environment Variables"
+                help="Place each environment variable on a new line, in the format <code>KEY=value</code>"
+                rows={3}
+              />
+              <div style={{ paddingLeft: 18 }}>
+                <input type="hidden" name="build[clean_checkout]" value="0" />
+                <label className="bold"><input name="build[clean_checkout]" type="checkbox" className="absolute" style={{ marginLeft: -18 }} value="1" />Clean checkout</label>
```
ticky: This checkbox’s selection outline gets occluded by the outer padding.
screen shot 2017-04-11 at 3 11 27 pm

app/lib/number.js
```diff
@@ -1,5 +1,7 @@
 import window from 'global/window';
 
+// Formats numbers nicely based on their current locale
+// e.g. 4200000 -> 4,200,000
```
ticky: oh, good idea!

app/components/pipeline/CreateBuildDialog.js
toolmantim: Oh, good catch! I also discovered that keyboard navigation was broke (you need to set tabIndex to -1 for the collapsed inputs), and some of the other outline and focus styles were broked. Now it should be both accessible to keyboard users and screenreaders, and that outline bug is fixed.

PR #238, opened 2017-04-11
New pipeline header layout improvements

toolmantim:
- [x] Figure out layout 
- [x] Add icons for other repo types
- [x] Fix smaller-screen layouts

### DIFF
--- a/app/components/pipeline/Header/index.js
+++ b/app/components/pipeline/Header/index.js
@@ -10,6 +10,7 @@ import CreateBuildDialog from '../CreateBuildDialog';
 import Builds from './builds';
 
 import permissions from '../../../lib/permissions';
+import { repositoryProviderIcon } from '../../../lib/repositories';
 
 class Header extends React.Component {
   static propTypes = {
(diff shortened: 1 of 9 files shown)


ticky:
Note: other repo provider icons have been added in #243!

toolmantim:
The design work for this is in App UI > Pipeline Header Layout Improvements

app/components/pipeline/Header/index.js
```diff
@@ -189,7 +174,7 @@ class Header extends React.Component {
     if (this.props.pipeline.description) {
       return <Emojify text={this.props.pipeline.description} />;
     } else {
-      return this.repositoryUrl();
+      return this.props.pipeline.repository.url;
```
ticky: this isn’t necessarily a web url, though!

toolmantim: Yeah, but this is just displayed as text if there's no description…

this is just shown here

ticky: Right, sorry, I meant for the hyperlinkification!

toolmantim: The `provider.url` will always be a web URL tho… if you check out the `Project::Provider::*` classes you can see they each have their own `project_url` methods, which is what we've been using until now to link to commits, pull requests, etc.

ticky: Huh, that’s weird, it didn’t change anything for me locally, always just output whatever git url was put in. Does that only work if you set up webhooks?

toolmantim: There's `repository.url` which is the raw value they put into the repository setting (which is what's printed under the name if it doesn't have a description), and then there's `repository.provider.url` which is generated from the Ruby's. It's used to detect the setup instructions too, to give the webhook setup screen. But none of that is affected by whether you setup webhooks or not… we don't know about their webhook setup really.

ticky: Hmm, not sure what I was thinking, then! :P

toolmantim: All good! As soon as I spotted that Javascript I was like "We should have that in Ruby already" but didn't get a chance to fully check it out.

These are the GraphQL models for it:

```ruby
module API::Graph::Types
  RepositoryType = GraphQL::ObjectType.define do
    name "Repository"
    description "A repository associated with a pipeline"

    field :url do
      type !types.String
      description "The URL for this repository"
      resolve -> (project, arguments, context) {
        project.repository
      }
    end

    field :provider, -> { API::Graph::Interfaces::RepositoryProviderInterface } do
      resolve -> (project, arguments, context) do
        Project::Provider.new(project)
      end
    end
  end
end
```

v.s.

```ruby
module API::Graph::Interfaces
  RepositoryProviderInterface = GraphQL::InterfaceType.define do
    name "RepositoryProvider"

    field :name, !types.String do
      resolve -> (provider, arguments, context) do
        provider.project.provider.name
      end
    end

    field :url, types.String do
      resolve -> (provider, arguments, context) do
        provider.project.provider.project_url(provider.project)
      end
    end
  end
end
```

`project.repository` is just the attribute from the ActiveRecord model.

`provider.project.provider.project_url` is the `#project_url` method of all the `Project::Provider::*` subclasses

toolmantim: We should probably improve the documentation of those GraphQL fields, because I think that wouldn't have helped with the confusion at all! I'll do that now.

ticky: Yeah, I just had a dig, and now I can’t see any way it could’ve been producing the behaviour I thought I’d observed. Maybe it was because it accepted either URL?

toolmantim: Yeah, I think because it was using `repositoryUrl()`

toolmantim:
Okay, this is ready to rumble. Thanks so much for your help @ticky, the new responsiveness is amaze.

PR #240, opened 2017-04-12
Limit job listing to a much smaller subset of jobs

toolmantim:
Currently the "all" query kills the DB, so we'll restrict the state list to active jobs (a much-smaller subset than all historical jobs") and we'll default it to show "scheduled" jobs. If we fix the performance problems, we can revert this back


keithpitt:
Nice!

PR #242, opened 2017-05-01
Improve Dialog behaviour

ticky:
This allows the dialog to scroll. What it does is rearrange the dialog to allow the entire dialog to appear partially offscreen, and the rest to scroll into view.

animated 2017-05-01 at 13 38 54

This has a couple of caveats, which might be acceptable or might need working around.

In any case, this is intended in support of #222, which can in some cases create dialogs which appear partially offscreen, and currently can't be scrolled to.

The caveats;
* Scrolling the dialog requires the cursor to be over the dialog, not the dialog backdrop
  * This is because the scroll container is transparent so the scroll event is passed to the backdrop, which is not part of a scroll container, and thus passed to the page.
  * We can work around this by putting the backdrop in the scroll container, but we need to think more about how we manage the transition animations if we do that.
* Firefox collapses the padding at the bottom of the flex container.

The easiest way to check out this behaviour is to open the Support dialog and shrink the window to ~320px wide and ~very smol high!

### DIFF
--- a/app/components/layout/Navigation/support-dialog.js
+++ b/app/components/layout/Navigation/support-dialog.js
@@ -59,7 +59,7 @@ class SupportDialog extends React.PureComponent {
   render() {
     return (
       <Dialog isOpen={this.props.isOpen} onRequestClose={this.props.onRequestClose}>
-        <div className="center" style={{ padding: "50px 10px" }}>
+        <div className="center" style={{ padding: "10% 2%" }}>
           {/* fyi the h1 class here is only necessary so this doesn't break on Bootstrap pages */}
           <h1 className="bold h1 mt0 mt2 mb4">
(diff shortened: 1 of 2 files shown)


ticky:
Yeah, that’s one of the caveats I mentioned. It needs further refactoring to make scrolling not suck. And _that_ requires figuring out how to do that without breaking the animations we’ve got going on.

ticky:
I’ve updated this with a fix for the Firefox/Safari caveat (added an element which gives us a faux padding at the bottom!)

toolmantim:
Nice fixes! ✈️

ticky:
This has been updated to lock the body’s scrolling when the dialog is present, which makes the noted caveat about cursor position acceptable as an interim improvement, as it’s impossible to scroll the _page_ in this way!

PR #248, opened 2017-05-05
Fix long menu items being cut-off on small screens

toolmantim:
Fixes:

before

To look like:

after

### DIFF
--- a/app/components/shared/Menu/button.js
+++ b/app/components/shared/Menu/button.js
@@ -24,7 +24,7 @@ class Button extends React.Component {
         link={this.props.link}
       >
         <div className="flex">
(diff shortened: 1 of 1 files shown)


ticky:
I was sure I made this change somewhere already 🙃

app/components/shared/Menu/button.js
```diff
@@ -24,7 +24,7 @@ class Button extends React.Component {
         link={this.props.link}
       >
         <div className="flex">
-          <div className="flex-auto">{this.props.children}</div>
+          <div className="flex-auto truncate">{this.props.children}</div>
```
ticky: oh right, there’s still a truncate on `BaseButton`, which isn’t doing anything - remove that!

PR #249, opened 2017-05-05
Make Panel.Header use a h2

toolmantim:
In general it's good to use semantic tags for accessibility, but not for styling, so let's use a h2 here for panel headers.

### DIFF
--- a/app/components/shared/Panel/index.js
+++ b/app/components/shared/Panel/index.js
@@ -6,6 +6,7 @@ import IntroWithButton from './intro-with-button';
 import Row from './row';
 import RowActions from './row-actions';
 import RowLink from './row-link';
+import Header from './header';
 
 class Panel extends React.Component {
   static propTypes = {
@@ -39,9 +40,9 @@ Panel.IntroWithButton = IntroWithButton;
 Panel.Row = Row;
 Panel.RowActions = RowActions;
 Panel.RowLink = RowLink;
+Panel.Header = Header;
 
 const SIMPLE_COMPONENTS = {
-  Header: 'bg-silver py2 px3 semi-bold',
   Footer: 'py2 px3',
   Section: 'm3'
 };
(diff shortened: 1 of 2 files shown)


app/components/shared/Panel/index.js
```diff
@@ -39,9 +40,9 @@ Panel.IntroWithButton = IntroWithButton;
 Panel.Row = Row;
 Panel.RowActions = RowActions;
 Panel.RowLink = RowLink;
+Panel.Header = Header;
 
 const SIMPLE_COMPONENTS = {
-  Header: 'bg-silver py2 px3 semi-bold',
```
ticky: I’d really like to make it so such element changes don’t require us to be splitting these back out into files like this. I was thinking about this today while I was doing #230 :(

toolmantim: I figured it was inevitable due to the combination of linter rules not allowing multiple components in a file, and the general inflexibility of this type of code: 

```
const SIMPLE_COMPONENTS = {
  Description: 'dark-gray mt1',
  Icon: 'flex items-center',
  Menu: 'flex items-center'
};

Object.keys(SIMPLE_COMPONENTS).forEach((componentName) => {
```

toolmantim: Just to clarifiy, I'm talking about how this code assumes a component only has two things: an identifier, and a list of class names.

Whereas if we just always used `React.PureComponent` for things, and made them declarative in JSX, then everything in the system is the same, and it's easy for anyone to, say, change a tag name, or a class name, or add an event handler, etc. And if we want to DRY some logic into a single place, we do that using our own functions calls from the components?

I just always find I regret defining classes/functions/things imperatively in loops like this, as it just makes it hard for anyone following me to fix/customize/adapt things.

toolmantim: I'm going to merge this, but I'm keen to hear your thoughts still!

PR #251, opened 2017-05-05
Use the mutation API directly in the Team Pipeline Chooser dialog

keithpitt:
Relay.Mutation does totes weird things when you use the autocomplete on the team show a bunch. When ever you search for a pipeline, it internally creates a new "relay connection" it's cache. If you search for 5 different things, it'll create 5 different cache entires.

Now when you add a new pipeline a team, Relay will try to refresh those 5 relay connections, which results in invalid mutations.

This change bypasses the auto-magic Relay mutation gear and just use the underlying API which doesn't do auto-refreshing of cache connection blah blah stuff and things.

### DIFF
--- a/app/components/team/Pipelines/chooser.js
+++ b/app/components/team/Pipelines/chooser.js
@@ -15,6 +15,7 @@ class Chooser extends React.Component {
 
   static propTypes = {
     team: PropTypes.shape({
+      id: PropTypes.string.isRequired,
       slug: PropTypes.string.isRequired,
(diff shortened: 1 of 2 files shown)


keithpitt:
@toolmantim this fixes that bug we were looking at yesterday!

toolmantim:
@keithpitt I think we need to update the `TeamMemberCreate` as well, because it was having similar problems?

keithpitt:
Ah right, will fix that up as well.

PR #253, opened 2017-05-05
Rename settings sections

toolmantim:


app/components/pipeline/SettingsMenu.js
```diff
@@ -57,7 +57,7 @@ class SettingsMenu extends React.Component {
         allowed: "pipelineUpdate",
         render: (idx) => (
           <Menu.Button key={idx} href={`${url}/queues`}>
```
keithpitt: Need to update the route here as well

PR #254, opened 2017-05-05
Bolder schedules icon

toolmantim:
Makes the schedules icon a similar darkness to the rest of the icons:

schedules


toolmantim:
(looks better with the proper `build-skipping` updates in #253)

PR #255, opened 2017-05-05
Give the YAML CodeMirror a min-height

toolmantim:
It's too small for tiny pipelines.


toolmantim:
I don't think this needs a review.

PR #256, opened 2017-05-05
Use the GraphQL mutation API directly when adding a user to a team

keithpitt:
Same as https://github.com/buildkite/frontend/pull/251 but for the users section


keithpitt:
@toolmantim fyi

PR #257, opened 2017-05-06
Fix menu button icon collapsing

toolmantim:
#248 introduced a bug where icons would be collapsed, and another where some menu items weren't properly truncated:

boogs

This fixes the two bugs by fixing the `flex` definitions. Rather than copy+pasting class names to every `<Icon>` the label and icon have been moved to props of the `<Menu.Button>`

fixed

This also removes the empty `flex-none` div in menu buttons without badges, and brings the icons a little closer to their labels.


toolmantim:
Going to merge this on green to get the bugfix out.

PR #258, opened 2017-05-08
Component to show build annotations

keithpitt:


app/graph/schema.json
```diff
@@ -10275,6 +10333,311 @@
...
+          "interfaces": [
+
+          ],
+          "enumValues": null,
+          "possibleTypes": null
+        },
+        {
+          "kind": "OBJECT",
+          "name": "Annotation",
+          "description": "A annotation on a build",
```
ticky: this is uh, not a helpful description - what on earth is an annotation?

keithpitt: Oh I haven't filled out the descriptions properly yet. I'll make sure not to copy/paste stuff from other parts before shipping :) Thanks for pointing it out! 👌

PR #261, opened 2017-05-11
Better Pusher connection handling

ticky:
This pull request includes a couple of little quality of life improvements for our push events.

1. When Pusher is disconnected, we’ll show the user a notice that things won’t live-update. In the past the page would just silently not update. Hopefully this’ll make it easier to tell the difference between a connection blip and a real showstopper!
   screen shot 2017-05-11 at 1 33 41 pm
   The notice is automatically dismissed when the connection is restored, and the user may dismiss it. Additionally, dismissing the notice will mean no new notification is created until the connection is restored at least once (this prevents notices _re_-appearing when going from `unavailable` -> `connecting` -> `unavailable`)
2. `NODE_ENV=development` Pusher logging is improved. Previously some logs were not shown as the logger wasn’t initialised in time. This is fixed, and we’re also logging more (read: all) Pusher connection state transitions.

### DIFF
--- a/app/components/layout/Flashes/index.js
+++ b/app/components/layout/Flashes/index.js
@@ -1,10 +1,13 @@
 import React from 'react';
 
 import FlashesStore from '../../../stores/FlashesStore';
+import PusherStore from '../../../stores/PusherStore';
 
 import Flash from './flash';
 
-class Flashes extends React.Component {
+const FLASH_CONN_ERROR_ID = 'FLASH_CONN_ERROR';
+
+class Flashes extends React.PureComponent {
   state = {
     flashes: []
   };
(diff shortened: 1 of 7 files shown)


app/components/layout/Flashes/index.js
```diff
@@ -16,18 +20,77 @@ class Flashes extends React.Component {
...
+  handleStoreChange = (payload) => {
+    this.setState({ flashes: this.state.flashes.concat(payload) });
+  };
+
+  hasConnectionErrorFlash() {
+    return this.state.flashes.length > 0 && this.state.flashes[0].id === FLASH_CONN_ERROR_ID;
   }
 
+  // show a flash when there's a push connection issue
+  handleConnectionError = () => {
```
keithpitt: Without knowing too much about when pusher sends the `unavailable` call, but perhaps we should maybe only show the flash if we've been unavailable for more than 10 seconds or so? Maybe more than 60 seconds? I just fear that Australia's shitty internet connect will mean that we're forever seeing that message.

ticky: Pusher only sends this event after the connection times out (30 seconds) plus it fails to connect once (10 seconds), so it can only happen once in a 40 second period.

keithpitt: Ah cool, good to know! Maybe a sneaky comment in there?

PR #264, opened 2017-05-15
Convert to Relay Modern (but not really)

sj26:
Lays the groundwork for migrating to React Modern. From the [conversion playbook](https://facebook.github.io/relay/docs/conversion-playbook.html), this switches to the dev version of react-relay and converts existing imports to `react-relay/classic`. The only gnarly piece is there's an import within react-router-relay outside our control, so I forked it and [released a version](https://github.com/sj26/react-router-relay/releases) which uses the classic import.

I'll keep this up to date until relay 1.0.0 final is released, then we can switch. Then comes the hard part: updating to use relay compat, adding a query renderer, ripping out react-router-relay, then switching to relay modern proper.

Things to check (from @keithpitt):
- [x] pipeline pages loads in dynamically OK
- [x] go to agents page, refresh, then go to pipelines page
- [x] make sure teams sections work (create mutations, etc)
- [x] make sure relay preloading still works (go to agents page and keep an eye on XHR requests)
- [x] add a "sleep 5" to our backend internal graphql endpoint and double check stuffs


sj26:
I've had a pretty good poke around after the upgrade, checked all the things in @keithpitt's list, and all seems well. Maybe good to go?

sj26:
You folks want to have a play as well and make sure it's all working?

PR #265, opened 2017-05-15
Team list in member edit

ticky:
- [x] When teams are added and removed the team chooser collection should be re-fetched
- [x] Teams editor shouldn't be displayed for organizations without teams enabled


ticky:
This now includes @sj26’ changes as a base! 😃

app/components/member/Edit/role.js
```diff
@@ -0,0 +1,147 @@
...
+          ? <span className="dark-gray">You can’t edit your own roles</span>
+          : (
+            <Button
+              onClick={this.handleUpdateOrganizationMemberClick}
+              loading={this.state.updating && 'Saving…'}
+            >
+              Save
+            </Button>
+          )
+      );
```
sj26: This is something I meant to look into: I don't think this is ever shown, currently. `allowed` is false for your own membership record. Maybe we need to switch this when `code` is `not_self` or something?

app/components/member/Edit/index.js
```diff
@@ -53,22 +50,21 @@ class MemberEdit extends React.PureComponent {
           }).isRequired
         ).isRequired
       }).isRequired
-    })
+    }),
+    relay: PropTypes.object.isRequired
```
sj26: Interesting, what does this do?

app/components/member/Edit/Memberships.js
```diff
@@ -27,6 +27,7 @@ class MemberEditMemberships extends React.PureComponent {
         ).isRequired
       }).isRequired
     }),
+    isSelf: PropTypes.bool.isRequired,
```
sj26: I couldn't figure out if this was state or a prop in my cut 😅 

I guess it's derived from props, so only changes when props change, so doesn't need to be tracked separately by the renderer as state? Does that mean isAdmin is also a prop, or something?

sj26:
Everything's looking great to me! Sorry for my newbie code, and thank you for integrating 🙇

app/components/member/Edit/role.js
ticky: Yeah, I think when I first built this I hadn’t realised that there was a permission with per-object details!

That said, the server-provided message is a bit ugly and awkward, so it might be worth changing.

app/components/member/Edit/index.js
ticky: Explicitly states you want access to `this.props.relay` so ESLint doesn’t scold you (it’s since moved to `member/Edit/memberships/index.js`, FYI!)

app/components/member/Edit/Memberships.js
ticky: Yours had it in the state, but in the end I moved it up the chain so the top-level element passes it to the children so they don’t all have to request `viewer` for that simple case. If they were to be used elsewhere they might need it again.

`isAdmin` remains state, and now lives in `member/Edit/role.js` (the only place it’s used)

ticky:
Updated with the following;

* Revised names for some components
* Adding organization members to teams

There's one giant caveat right now; we can't do a query for teams a member is not in!

sj26:
The team chooser now re-fetches every time the dialog is opened.

I tried to configure the TeamMemberDelete mutation to invalidate the organization.teams connection, but I couldn't figure out how to do it with userSelector in the mix.

(I think Relay Modern will make this easier, eventually.)

sj26:
This change also means that the chooser doesn't appear until the refetch is finished. It would be possible to add a "loading" state in the middle there, but it's pretty quick so I didn't bother. If you think it needs one then maybe add it in?

app/components/member/Edit/index.js
```diff
@@ -71,10 +88,7 @@ class Edit extends React.PureComponent {
             viewer={this.props.viewer}
             organizationMember={this.props.organizationMember}
           />
-          <MemberEditTeamMemberships
-            isSelf={isSelf}
-            organizationMember={this.props.organizationMember}
-          />
+          {memberEditTeamMemberships}
```
ticky: Shouldn’t this be a responsibility of the `MemberEditTeamMemberships` component, rather than its parent?

ticky:
Yeah, I looked at this solution but this doesn’t solve for the fact that lists with a search query are not invalidated and thus if you search for the name of a team, add it, and then perform the same search, you get the old results.

ticky:
I’ve spent some time trying to find a better solution to this, and I really don’t know what it is. I suspect it’ll require some GraphQL magick, which perhaps only @keithpitt knows how to achieve. The Learn GraphQL slack didn’t have any answers, either.

app/components/member/Edit/index.js
sj26: Oh, good point!

sj26:
Oh, bummer. I didn't see that caveat. :-(

I'm pretty sure we need to teach the mutation how to `RANGE_DELETE` from `organization.teams` connections with a relevant `user: ...` argument, but I don't know how.

ticky:
Yeah, that’s the thing, I really can’t see how that’s supposed to work in GraphQL mutations, at all. The documentation is just so obtuse 😕

keithpitt:
Ah! So, `setVariables` will change the Relay variables - and talk to the Relay cache if it can.

`forceFetch` has exactly the same API, but it skips the cache. If you switch calls from `setVariables` to `forceFetch`, then the issue should be fixed!

sj26:
Oh, so maybe we need to `forceFetch` for the search queries, too? Seems less than ideal, but it would work!

keithpitt:
Yeah, the issue is when ever you search for something, it creates a new "relay connection" cache. Since it treats:

`team.pipelines()`

As a seperate connection from:

`team.pipelines(search: "f")`

And...

`team.pipelines(search: "fr")`

And...

`team.pipelines(search: "front")`

And...

`team.pipelines(search: "frontend")`

Etc...

ticky:
Yes but replacing it with `forceFetch` means every single keystroke will, after up to 100ms, send a request to the server which cannot be canceled, which seems potentially lethal on *this page* in particular.

If there’s no other proposed solution, then perhaps my slightly less brutal “keep a list of queries we’ve sent, `forceFetch` if they match” is actually reasonable? 😕

ticky:
Phew! That was quite an adventure!

PR #266, opened 2017-05-17
Push a final image to 🐳 buildkite/frontend on master

lox:
Currently a docker image gets built for frontend, but only pushed to [`buildkiteci/frontend`](https://hub.docker.com/r/buildkiteci/frontend/), which I view as ephemeral build artifacts and prone to being deleted. 

This change adds a push step at the end that deploys a passing image to the more official [`buildkite/frontend`](https://hub.docker.com/r/buildkite/frontend/) with tags that indicate what git commit they contain. For now, this will have no practical impact on anything, but it provides public runnable images of our frontend that we can also possibly consume in our development environments.

This also adds a `docker-compose.override.yml` which is only read in development (specifically, when no config files are specified). This mounts in the project for actually development using the new [`cached` directive](https://docs.docker.com/docker-for-mac/osxfs-caching/) which is much faster for Docker for Mac.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,14 +4,9 @@ services:
   frontend:
     image: index.docker.io/buildkite/frontend
     build: .
-    volumes:
-      - "./bundle-analysis:/frontend/bundle-analysis"
-      - "./dist:/frontend/dist"
-      - "./coverage:/frontend/coverage"
     environment:
-      BUILDKITE:
-      BUILDKITE_COMMIT:
-      BUILDKITE_ORGANIZATION_SLUG:
-      BUILDKITE_PIPELINE_SLUG:
-      CI:
-      EMOJI_HOST:
+      - EMOJI_HOST
+      - FRONTEND_HOST
+    volumes:
+      - ".:/frontend:cached"
+      - /frontend/node_modules
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,6 +7,11 @@ services:
     environment:
       - EMOJI_HOST
       - FRONTEND_HOST
+      - BUILDKITE
+      - BUILDKITE_COMMIT
+      - BUILDKITE_ORGANIZATION_SLUG
+      - BUILDKITE_PIPELINE_SLUG
+      - CI
     volumes:
       - ".:/frontend:cached"
       - /frontend/node_modules
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,6 +2,7 @@ version: '2'
 
 services:
   frontend:
+    image: index.docker.io/buildkite/frontend
     build: .
     volumes:
       - "./bundle-analysis:/frontend/bundle-analysis"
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -2,16 +2,16 @@ version: '2'
...
-      FRONTEND_HOST:
+      - EMOJI_HOST
+      - FRONTEND_HOST
+      - BUILDKITE
+      - BUILDKITE_COMMIT
+      - BUILDKITE_ORGANIZATION_SLUG
+      - BUILDKITE_PIPELINE_SLUG
+      - CI
+    volumes:
+      - ".:/frontend:cached"
```
toolmantim: woo what's this do?

lox: It's a very, very recent addition that allows docker for mac to make some tuning decisions on volume mounts from the host filesystem: https://docs.docker.com/docker-for-mac/osxfs-caching/

> cached: the host’s view is authoritative
> (permit delays before updates on the host appear in the container)

The TL;DR is that it makes anything that interacts with the filesystem much faster in the container at the expense of some eventual consistency with the host.

lox: I've been experimenting with what best practice is for volume mounts, I tend to default to `:cached,ro` if I can now and then have more granular mounts for things like logs or cache or whatever that will actually have things written to it by the container.

toolmantim: Interesting!

I went away from recommending everyone just mount `.` into the container, and instead just mount the directories that have artifacts. Just to stop the volume mount clobbering the standard `ADD/COPY` files in the container's fs. Hence https://github.com/buildkite-plugins/docker-compose-buildkite-plugin#artifacts. Perhaps going back to `.` is a better idea?

lox: Doesn't that cause problems because the ephemeral build scripts get written in the project root?

toolmantim: The docker-compose plugin doesn't use ephemeral scripts, because it overrides the `command` hook (which is the thing that generates and uses those ephemeral scripts).

lox: Oh yeah. I should know that.

toolmantim: So… given a world without emphemeral steps, and "normal docker" being to `ADD/COPY` files into the container, do we want to be recommending linking the `.` volume.

lox: No, I don't think so. I think the approach of just mounting in the directories with the artifacts is a good one.

toolmantim: Cool cool, I agree! I'll let you revert it back.

lox:
@toolmantim after some thought, I moved the mount of `.` into the `docker-compose.override.yml`, which I generally see as the "things I want in dev, but not CI" file.

lox:
Any objections to pushing a public image of our frontend code @toolmantim @ticky? I figure seeing as dev is already done in the open it wouldn't be controversial.

lox:
Eeep. Hopefully not, because it just happened :)

PR #267, opened 2017-05-18
Paginate main Pipelines page

ticky:
This reduces the initial pipeline fetch to 100 pipelines, and then makes each subsequent page a 50-pipeline chunk.

We've had some feedback that the graphs are causing all sorts of slowdown for larger customers, and the only simple way to fix that would seem to be to make it less demanding overall!

should help with buildkite/feedback#247

### DIFF
--- a/app/components/organization/Pipelines.js
+++ b/app/components/organization/Pipelines.js
@@ -2,13 +2,18 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Relay from 'react-relay';
 
+import Button from '../shared/Button';
 import SectionLoader from '../shared/SectionLoader';
+import Spinner from '../shared/Spinner';
 
 import FlashesStore from '../../stores/FlashesStore';
 import UserSessionStore from '../../stores/UserSessionStore';
 
 import Pipeline from './Pipeline';
 import Welcome from './Welcome';
+
+const INITIAL_PAGE_SIZE = 100;
+const PAGE_SIZE = 50;
 
 class OrganizationPipelines extends React.Component {
   static propTypes = {
@@ -22,7 +27,10 @@ class OrganizationPipelines extends React.Component {
               id: PropTypes.string.isRequired
             }).isRequired
           }).isRequired
-        )
+        ),
+        pageInfo: PropTypes.shape({
+          hasNextPage: PropTypes.bool.isRequired
+        }).isRequired
       })
     }),
     relay: PropTypes.object.isRequired,
@@ -122,6 +130,7 @@ class OrganizationPipelines extends React.Component {
       return (
         <div>
           {this.renderPipelines()}
+          {this.renderPipelineFooter()}
         </div>
       );
     } else {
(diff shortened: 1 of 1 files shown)


ticky:
Further (out of scope for this PR) optimisations I can see from here;

* Fetch starred pipelines first (or separately) with graphs, along with unstarred sans graphs, then graphs for all!
* Load graph data per-pipeline as you scroll

toolmantim:
This is exciting!

ticky:
One thing that might be worthwhile to check is how many of the customers who’ve had perf issues have more than 100 pipelines - that may still be too generous an initial page size!

sj26:
This will definitely be an improvement though. I'd say ship it and tweak.

toolmantim:
Dumb question: does this handle pipeline favoriting?

ticky:
@toolmantim it handles it in the same way we currently handle it; we hope that favourited pipelines are in the first response - if they’re not, they’ll show up there in subsequent paginations.

I do plan to implement the groundwork to make them filterable, but I think this is a reasonable compromise for the moment.

toolmantim:
I was going to suggest that between pagination and favouriting, we probably don't need filtering! That is, if we made sure that we always load favourites…

toolmantim:
Just to confirm, if you had a favourite pipeline in the 4th page of results, would this mean it wouldn't appear at the top, until you scrolled to the bottom and hit "Load more pipelines…" 4 times?

toolmantim:
Or do favourites get sorted on the backend to return at the top of the results?

ticky:
That’s correct, it will appear at the top only _after_ pagination. Currently, if you have more than 500 pipelines, you can’t see the rest. Tradeoffs!

Filtering would still be good, as there are issues around favouriting which make it difficult to support from a GraphQL perspective.

From Slack;
> so I was going to update this with backend changes to let pipelines be filtered by `favourited` status but it raises questions about how many favourites to _expect_ and/or show
>
> so I think leaving it like this, with a better balance of caveats, is going to be better

ticky:
For instance, for a client with, say, 650 pipelines, this trades not having all the pipelines immediately visible for “you can’t ⌘F for a pipeline on page load”

toolmantim:
So if I'm a person at REA, and I have 5 favourited pipelines (out of 400) scattered throughout the alphabet, I'd have to load all 400 pipelines to see my favourites up the top? Would I even know to look up the top when the page is scrolled down the bottom?

Perhaps we can fix this by simply changing the backend sorting algorithm to "Favourites A-Z, then non-favourites A-Z"?

sj26:
Or just load all favourites all the time up the top in a separate component?

toolmantim:
Yeah, that could work too!

ticky:
Loading a sorted list would actually solve several of the questions about how best to delineate the two lists. Just having the favourites returned first means if you have >100 favourites (unusual, yes, but possible) it just paginates correctly!

ticky:
That said it limits us if we were ever to provide that “sort by last activity” view people keep asking for 👀

PR #268, opened 2017-05-18
Improve Schedules Build form labels and layouts

toolmantim:
Some cleanups before we announce scheduled builds are live:

* Clarifies which fields are required
* Puts the description at the top before all the Hard Stuff
* Added more information about the the cron interval syntax, along with `@` extensions we support
* Made sure teams forms were consistent

There's some things that still need improving which people might ask:

* Such as "How do I change/update the "Creator" of the builds?"
* What happens if the "Creator" is removed from the org?

Before:

before-1

After:

after-1

Before:

before-2

After:

after-2

Before:

before-3

After:

after-3

### DIFF
--- a/app/components/pipeline/schedules/Form.js
+++ b/app/components/pipeline/schedules/Form.js
@@ -49,25 +49,28 @@ class Form extends React.Component {
 
         <FormTextField
           label="Message"
-          help="The message to use for the build. Default: “Scheduled Build for (Pipeline Name)”"
+          help="The message to use for the build."
           errors={errors.findForField("message")}
           value={this.props.message}
+          placeholder="Scheduled build"
           ref={(messageTextField) => this.messageTextField = messageTextField}
         />
 
         <FormTextField
           label="Commit"
-          help="The commit to use for the build. Default: “HEAD”"
+          help="The commit ref to use for the build."
           errors={errors.findForField("commit")}
           value={this.props.commit}
+          placeholder="HEAD"
           ref={(commitTextField) => this.commitTextField = commitTextField}
         />
 
         <FormTextField
           label="Branch"
-          help={`The branch to use for the build. Default: the pipeline’s default branch (currently “${this.props.pipeline.defaultBranch}”)`}
+          help="The branch to use for the build. Defaults to the pipeline’s default branch."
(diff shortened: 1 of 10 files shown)


app/components/pipeline/schedules/Form.js
```diff
@@ -30,50 +30,50 @@ class Form extends React.Component {
...
-          help="Describe what this schedule is all about (you can even use :emoji:)"
-          errors={errors.findForField("label")}
-          value={this.props.label}
-          ref={(labelTextField) => this.labelTextField = labelTextField}
-        />
-
-        <FormTextField
           label="Message"
-          help="The message to use for the created build"
+          help="The message to use for the build. Default: “Scheduled Build for (Pipeline Name)”"
```
harrietgrace: Does that default auto-populate? Because it wasn't appearing in the screenshots of the hourly build?

app/components/pipeline/schedules/Form.js
```diff
@@ -30,50 +30,50 @@ class Form extends React.Component {
...
           errors={errors.findForField("commit")}
           value={this.props.commit}
-          placeholder={"HEAD"}
           ref={(commitTextField) => this.commitTextField = commitTextField}
         />
 
         <FormTextField
           label="Branch"
-          help="The branch to use for the created build"
+          help={`The branch to use for the build. Default: the pipeline’s default branch (currently “${this.props.pipeline.defaultBranch}”)`}
```
harrietgrace: Niiiice 👌🏻 What do you think about saying 'yours is currently __'? Otherwise people might think this help text is generic?

toolmantim:
That's awesome feedback @harrietgrace! I've just updated it again, and moved it back to using placeholders:

after-5

What do you reckon?

ticky:
“Placeholders as defaults” feels like a strong UI move we’ve made, so good to keep it consistent!

toolmantim:
@ticky oh yeah. Strong in this PR, or that's a general pattern we already use? We ended up deciding just to prefill the fields instead.

ticky:
It’s a pattern we already use! The new builds popup and such all show their fallback when empty! It’s a good use of the placeholder, IMO!

PR #269, opened 2017-05-18
Hide invite teams without teams

sj26:
While investigating a flakey spec I noticed the invite page shows a _Teams_ section even if the organization doesn't have teams enabled:

image

This just checks the teams count and hides teams if there aren't any at all, which should be a reasonable check because there should at least be an _Everyone_ team at the moment, and at least one team otherwise:

image

Which reminds me, we should probably make sure there is always at least one team in an organization.


ticky:
Oops! Good catch! 😄

PR #270, opened 2017-05-19
Update support for Emoji 4.0 (Unicode 9.0)

ticky:
frontend counterpart to buildkite/emojis#67

PR #273, opened 2017-05-24
Emoji 4.0 support & only preprocessing emoji shortcodes

ticky:
Reverts buildkite/frontend#272, restoring buildkite/frontend#270.


app/css/emoji.css
```diff
@@ -1,5 +1,13 @@
...
-  width: 1.25em;
-  height: 1.25em;
+  width: 1.23em;
+  height: 1.23em;
+  margin-left:.05em;
+  margin-right:.05em;
   vertical-align: -.2em;
 }
+
+.tiny-kitemoji .emoji {
```
keithpitt: Kitemoji! Possible the best name ever?

app/components/shared/Emojify.js
```diff
@@ -3,25 +3,35 @@ import PropTypes from 'prop-types';
 
 import Emoji from '../../lib/Emoji';
 
-class Emojify extends React.PureComponent {
+export default class Emojify extends React.PureComponent {
```
sj26: Making this pure seems like it could speed things up a bit :clap:

app/lib/Emoji.js
```diff
@@ -1,78 +1,49 @@
...
-  }
-
-  _image(catalogue, emoji) {
-    // Emoji catalogue hosts have a normalized host that always end with a "/"
-    const emojiUrl = `${catalogue.host}${emoji.image}`;
-    const emojiCanonicalRepresentation = emoji.unicode || `:${emoji.name}:`;
+      return match;
 
-    return `<img class="emoji" title="${emoji.name}" alt="${emojiCanonicalRepresentation}" src="${emojiUrl}" draggable="false" />`;
+    });
```
sj26: Nice. :100:

One of the best performance tweaks I made to the backend emoji processing code was inverting the loops, changing to a single scan over the text which then looped through the catalogues to find matches. Is that something which might help here? I know they behave slightly differently here but that behaviour could be combined.

ticky: It sure could be, yeah!

ticky:
Bonus points: this now cuts the client-side Apple emoji catalogue size down by about ⅓ by removing the unicode version of the index, as well as all the image URLs (as they’re no longer used). We could go further if we’re convinced we’re not going to be landing that custom emoji picker anytime soon. :wink:

ticky:
The bug I was trying to mitigate by adding server-side support to this showed up in testing. After a bit of research, I ended up landing ticky/mojibaka#8, which allowed the tests to be moved to happen immediately in app.js’ context, rather than after the entire page loads, this seems to mitigate it properly!

PR #274, opened 2017-05-24
Improved pipeline filtering

ticky:
This adds a filtering option to the Pipelines list.

The header has been made responsive, so the search field will wrap onto a new line at small resolutions.

screen shot 2017-05-24 at 3 17 45 pmscreen shot 2017-05-24 at 3 18 06 pm

Filters are added to the URL, meaning you can navigate back to them using the browser’s native back button.

A new empty state has been added for when searching.

screen shot 2017-05-24 at 3 20 40 pm

Note that this also changes `SearchField`s’ placeholder colour to match their text colour and icon colour, and adds a `borderless` variant to them, which disables borders and inner shadows.


toolmantim:
There seems to be a bug with the "Load more pipelines…" for me, it reloads the whole pipelines list:

loading-bug

Not sure if that's just my machine?

toolmantim:
Also another bug I just spotted: if you click "Pipelines" in the top-level org nav it doesn't clear the filter (but it does clear the URL param)

toolmantim:
Okay, I've just pushed up a bunch of things I was able to improve… all pretty small.

I noticed the search results are a little slow… I think because they're loading in all the graph data as well. And it'd be great if we could get the favourites loading at the top somehow!

SomeoneWeird:
🙏

keithpitt:
This is rad! @ticky when we chat about the other Relay stuff tomorrow we can talk about a cunning plan I have to make this page even faster! 🐎

But in the mean time, I think this is awesome to ship to start solving problems right away.

app/components/shared/Icon/search.svg
```diff
@@ -1 +1,2 @@
-<path d="M19.58,17.61l-6.29-6.13a6.79,6.79,0,0,0,1.36-4.07,1.35,1.35,0,0,0,0-.23,1.36,1.36,0,0,0,0-.23A7,7,0,0,0,7.53,0L7.33,0,7.12,0A7,7,0,0,0,0,6.95a1.33,1.33,0,0,0,0,.23,1.33,1.33,0,0,0,0,.23,7,7,0,0,0,7.12,6.95l0.2,0,0.2,0a7.2,7.2,0,0,0,3.64-1l6.38,6.22a1.46,1.46,0,0,0,2,0A1.38,1.38,0,0,0,19.58,17.61Zm-12.25-6-0.2,0A4.2,4.2,0,0,1,2.87,7.41a1.32,1.32,0,0,0,0-.23,1.33,1.33,0,0,0,0-.23A4.2,4.2,0,0,1,7.12,2.81l0.2,0,0.2,0a4.2,4.2,0,0,1,4.25,4.14,1.35,1.35,0,0,0,0,.23,1.36,1.36,0,0,0,0,.23,4.2,4.2,0,0,1-4.25,4.14Z"/>
+<circle cx="8.0503" cy="8.0503" r="6" transform="translate(-3.3345 8.0503) rotate(-45)" style="fill:none;stroke:currentColor;stroke-miterlimit:10;stroke-width:2px"/>
+<line x1="12.2929" y1="12.2929" x2="17.9497" y2="17.9497" style="fill:none;stroke:currentColor;stroke-linecap:round;stroke-miterlimit:10;stroke-width:2px"/>
```
ticky: yessss, thank you, this is much nicer! 😃

ticky:
Latest commits fix [this bug](https://github.com/buildkite/frontend/pull/274#issuecomment-303920393)!

ticky:
@keithpitt @toolmantim (RE: your discussion on Slack) I was looking into this and I think using a custom sort (favourited, then by name) is going to be cleaner than doing anything else; it reduces the complexity of this component (as it means we don’t need to manually separate the lists on every render) and means that if someone exceeds `PAGE_SIZE` favourites, we don’t need to do any special handling at all!

ticky:
> Also another bug I just spotted: if you click "Pipelines" in the top-level org nav it doesn't clear the filter (but it does clear the URL param)

Ugh, of course, it would only work if this were a managed component, and of course, we don’t really want it to be. Shit.

ticky:
Okay, so I’ve switched up the search input to manage its values (which is a shame but it’s a pretty lightweight hack!) and now it’s good.

ticky:
Regarding query slowness, it’s not too bad for me locally, but I know we’ve gone with this query approach because we wanted a single query to fire off to fetch the graph data, but I think we might need to find a way to paginate that, somehow - it’s just too intensive.

toolmantim:
@ticky ah, with the slowness I was just testing with 750 pipelines where a search would match at least a full `PAGE_SIZE` worth of pipelines, and it felt a little slow. But we can worry about that later after testing in prod too!

toolmantim:
Whoops, wrong button :)

PR #275, opened 2017-05-25
Work around webpack-dev-server hostname checking requirement

lox:
To fix security things, webpack-dev-server added a check for hostname in https://github.com/webpack/webpack-dev-server/issues/887. 

Unfortunately, they also used that host check as the address to bind to, with only 127.0.0.1 and localhost as exceptions. This breaks docker containers or proxies. 

The workaround is to set up a host entry in the container for `buildkite.dev` to point to `0.0.0.0`. This then still has the hostname check, but it binds to the correct interface in the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,4 +7,5 @@ ADD package.json yarn.lock /frontend/
 RUN yarn install
 
 ADD . /frontend/
+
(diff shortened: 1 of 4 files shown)


lox:
Insta-merging given I'm the only consumer of these docker containers.

PR #276, opened 2017-05-25
Add a "reset" event to FlashesStore

ticky:
This facilitates us clearing the displayed flashes on navigation


sj26:
Oh neat! Are we displaying them multiple times by mistake?

I think the idea of flashes from a rails perspective is that they evaporate when they are shown. Should we be removing them once they've been displayed? But I guess removing on navigation should work, too?

ticky:
Define “multiple times?”

They currently have no way to automatically hide because a) the `FlashesStore` is not a store, more a “notifier,” and b) the `Flashes` component is what keeps track of active flashes, and it has no way to remove them other than user interaction. This solves b without rearchitecting a.

ticky:
For reference, you can compare behaviours by flashing some random messages locally;

```javascript
((FlashesStore) => (
  FlashesStore.flash(FlashesStore.INFO, "Something’s gone terribly wrong!")
))(Webpack.require('stores/FlashesStore'))
```

Compare navigating between Rails pages and SPA pages; Rails pages will cause the Flashes component to re-render with no flashes, whilst SPA pages will not, until this patch!

PR #282, opened 2017-05-28
Webpack tweaks

sj26:
Two changes:

Tell babel to pass through modules so that webpack is fully away of imports/exports which should make hot reloading better and enable switching away from our deprecated react-hmre preset to react-hot-reloader.

And embrace the "latest" es standards not just 2015 but 2016 and 2017, too. Async awaits!


sj26:
Popped a couple more bits on.

Moves some of our package.json dependencies which are only for dev into dev dependencies.

And remove some now-unused metadata from the package.json.

.babelrc
```diff
@@ -1,5 +1,9 @@
 {
-  "presets": ["react", "es2015", "stage-0"],
+  "presets": [
+    "react",
+    ["latest", {"es2015": {"modules": false}}],
+    "stage-0"
+  ],
```
ticky: [`latest` is actually deprecated](http://babeljs.io/docs/plugins/preset-latest/) in favour of using [`env`](http://babeljs.io/docs/plugins/preset-env/), which, if we’re changing it anyway, I think we should use instead!

It lets us frame the compilation in regards to browser compatibility rather than “versions” of JavaScript!

.babelrc
```diff
@@ -8,7 +12,10 @@
       "presets": ["react-hmre"]
     },
     "test": {
-      "plugins": ["react-pure-components"]
+      "plugins": [
+        "react-pure-components",
+        "transform-es2015-modules-commonjs"
```
ticky: What’s this adding?

package.json
```diff
@@ -5,11 +5,6 @@
     "type": "git",
     "url": "https://github.com/buildkite/frontend.git"
   },
-  "metadata": {
-    "graphql": {
-      "schema": "./app/graph/schema.json"
-    }
-  },
```
ticky: What happened to this? Was this removed in relay modern?

sj26: Ah, I think the explains were in the commit title :-)

> This was for babel-relay-plugin-loader

.babelrc
sj26: Oh neat, I did see the env present when I was researching webpacker but latest seems closer to es2015 (it's just a superset), I didn't realise it was deprecated, and I didn't want to do too much in one change.

Maybe we could use latest in this PR and open another for env?

sj26: Oh no, I see that [it defaults to the same as latest](http://babeljs.io/docs/plugins/preset-env/#how-it-works-support-all-plugins-in-babel-that-are-considered-latest). I'll switch to env!

.babelrc
sj26: Ah, I think this got lost in [the commit](https://github.com/buildkite/frontend/pull/282/commits/8b4c8d12a3d616f6f77a106b75c0627eca6dab36) title again, sorry:

> Jest still needs commonjs modules
> 
> This is what the docs recommend:
> 
> https://facebook.github.io/jest/docs/webpack.html#using-with-webpack-2

ticky: Hmm, should we confine it to the `test` environment, then, as it suggests?

ticky: Ugh, sorry, clicked the wrong link. All good! 😃

PR #284, opened 2017-05-29
Modernise pipeline teams ui

ticky:


ticky:
This probably needs a `pipelineSelector` on `organization.teams` added in buildkite/buildkite so some of the complexity can be reduced. Might also be a good time to think about server-side permission-based filtering?

PR #288, opened 2017-05-31
Adding Relay Compiler

keithpitt:
The aim of this PR is to compile Relay queries, store them as artifacts...and do _something_ with them...

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,7 +5,7 @@
     "stage-0"
   ],
   "plugins": [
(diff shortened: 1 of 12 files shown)


sj26:
The relay compiler just needs to run right before the webpack build. If you install `relay-compiler-webpack-plugin` and add it to the webpack config then it will run it automatically, including during hot reloading. 👌

ticky:
Could you please add the `watchman` requirement to the setup docs?

PR #289, opened 2017-05-31
Ask GraphQL for agent priority, too

sj26:
We seem to have missed a field in the graphql query required for showing the agent priority on the new agent show page, so here's a tiny change to include it!

I wonder if we should be showing priority in the agent index as well?

### Before

image

### After

image


sj26:
It doesn't look like this info is preloaded anywhere so this seems to be enough.

PR #290, opened 2017-05-31
Replace function refs with reffer refs

ticky:
I totally forgot that I had published [`reffer`](https://www.npmjs.com/package/reffer) and had intended to get it in here, since string refs aren’t around anymore, and throwing weird functions around always bugged me!

### DIFF
--- a/app/components/agent/Index/installation.js
+++ b/app/components/agent/Index/installation.js
@@ -1,8 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Relay from 'react-relay/classic';
-import { seconds } from 'metrick/duration';
 import Confetti from 'react-confetti';
+import { seconds } from 'metrick/duration';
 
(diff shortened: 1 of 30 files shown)


toolmantim:
Interesting!

Looking at this code change made me wonder… is the real problem that we're using overly long arg names?

```js
ref={(buildMessageTextField) => this.buildMessageTextField = buildMessageTextField}
```

vs just:

```js
ref={(el) => this.buildMessageTextField = el}
```

vs referrer:

```js
ref={this::reffer("buildMessageTextField")}
```

Reffer is less characters, but it's also less clear that it sets `this.buildMessageTextField`, which is what you need if you want to access it. And what you need to make sure you don't clobber/shadow with some other function.

Hrmm…

ticky:
Doesn’t seem like there was much enthusiasm here, and that’s fair. Going to close this one!

PR #291, opened 2017-05-31
Reusable “Show More” component

ticky:
This implements a basic proof of the new Connection-as-interface stuff in all the places we previously had an ad hoc implementation of more or less the same pagination!

### DIFF
--- a/app/components/shared/ShowMoreFooter.js
+++ b/app/components/shared/ShowMoreFooter.js
@@ -1,9 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Relay from 'react-relay/classic';
+import classNames from 'classnames';
 
 import Button from './Button';
-import Panel from './Panel';
 import Spinner from './Spinner';
 
 class ShowMoreFooter extends React.PureComponent {
@@ -16,12 +16,14 @@ class ShowMoreFooter extends React.PureComponent {
     onShowMore: PropTypes.func.isRequired,
     label: PropTypes.string,
     loading: PropTypes.bool,
-    searching: PropTypes.bool
+    searching: PropTypes.bool,
+    className: PropTypes.string
   };
 
   static defaultProps = {
     loading: false,
-    searching: false
+    searching: false,
+    className: 'px3 py2'
   };
 
   render() {
@@ -54,9 +56,9 @@ class ShowMoreFooter extends React.PureComponent {
     }
 
     return (
-      <Panel.Footer className="center">
+      <div className={classNames('center', this.props.className)}>
         {footerContent}
-      </Panel.Footer>
+      </div>
     );
   }
 }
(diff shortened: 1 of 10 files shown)


keithpitt:
Nice! How does it work with the new Relay Modern paginationContainer gear?

ticky:
Hmm, it sort of comes at this from a different direction. It should be easily adaptable to work within that framework, however, because currently some of our components implement two paginators within a single component, `PaginationContainer` isn’t really able to be dropped in to those until _that’s_ refactored.

I think ultimately the way forward will be to eventually expand this into a generic pagination container handler which sits atop the `PaginationContainer`, rather than expecting to move wholesale to it right now.

ticky:
Going to rethink this in the context of Relay Modern. Might be able to make this even nicer!

ticky:
Having looked at relay-modern, I think we're going to need more thought put into how we handle this. I believe this will end up being some sort of refactor to use paginationContainer directly, but as we're not currently able to use it as our API doesnt implement the heretofore unpublished updated pageInfo spec used in Relay Modern (which I don't know if relay modern has an issue open about and haven't posted for fear of being chewed out by Facebook devs) I think we should merge this sooner than later and think about that when we've landed relay modern compatibility in our API.

ticky:
That is to say I think this is an improvement even if it doesn't use relay modern and I'd rather clear this from the table before we look at that, because that requires widespread systemic change which would itself not break this change 👌🏼

app/components/shared/ShowMoreFooter.js
```diff
@@ -0,0 +1,76 @@
...
+        onClick={this.props.onShowMore}
+      >
+        Show more{this.props.label ? ` ${this.props.label}` : ''}…
+      </Button>
+    );
+
+    // show a spinner if we're loading more
+    if (loading) {
+      footerContent = <Spinner style={{ margin: 9.5 }} />;
+    }
```
sj26: Should this maybe be an if/else so we don't construct a Button just to throw it away? Although I'm not that familiar with how important that sort of churn is in js gc, especially considering these are abstract dom elements. 😅

sj26: (I miss ruby's implicit return value for things like conditionals in javascript 💖 )

ticky: I don’t really like Ruby’s implicit return values 🙅🏻 but I do kind of wish there was *a* way to return something from a conditional. (I think this is why I like ternary statements, to some peoples’ chagrin)

I don’t believe this approach has any major cost (all this will do is call “React.createElement”, which I don’t believe will end up doing anything meaningful until the element is actually mounted, which it thus won’t be.

ticky: Additionally, this is how the individual paginators work currently, so I don’t believe it’s hurting us too much, and now if we _do_ decide to optimise it further, we get to do it once :wink:

PR #294, opened 2017-06-07
Prevent Create Build dialog submit without required fields

toolmantim:
A small fix for this tweet:
https://twitter.com/RobStuttaford/status/870509925016813573


toolmantim:
I might just merge this, but I'm happy to fix it up based on your review @ticky

PR #295, opened 2017-06-08
Handle private GitHub emails in email prompt

ticky:
This has been a little improvement that feels like it should’ve been here forever, but here we go!

This updates the email prompt component to better deal with GitHub’s private email addresses. Given they’re predictable, instead of prompting the user to add the email address to their account…

screen shot 2017-06-08 at 4 15 27 pm

…let’s instead prompt them to link their GitHub account, which will _actually_ connect the builds and not require a weird email verification dance!

screen shot 2017-06-08 at 4 16 30 pm

“Manage Connected Apps” takes the user to `/user/connected-apps` where they can add their GitHub account. We _could_ hypothetically make this link directly do the GitHub auth flow, but this component doesn’t (read: currently can’t!) actually check if an account is connected, and it feels like it’d be weird to do that straight from a build page.

It also _doesn’t_ address the case of a user having an account connected already - either because this is actually another user’s GitHub account, or because they’ve changed account names and our cache hasn’t caught up.

### DIFF
--- a/app/components/build/AvatarWithUnknownEmailPrompt.js
+++ b/app/components/build/AvatarWithUnknownEmailPrompt.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Relay from 'react-relay/classic';
 
+import Button from '../shared/Button';
 import Dropdown from '../shared/Dropdown';
 import Icon from '../shared/Icon';
 import Spinner from '../shared/Spinner';
@@ -11,6 +12,8 @@ import FlashesStore from '../../stores/FlashesStore';
 
 import EmailCreateMutation from '../../mutations/EmailCreate';
 import NoticeDismissMutation from '../../mutations/NoticeDismiss';
+
+const GITHUB_NOREPLY_DOMAIN = '@users.noreply.github.com';
 
 class AvatarWithUnknownEmailPrompt extends React.PureComponent {
   static propTypes = {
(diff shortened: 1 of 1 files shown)


app/components/build/AvatarWithUnknownEmailPrompt.js
```diff
@@ -114,49 +117,90 @@ class AvatarWithUnknownEmailPrompt extends React.PureComponent {
       return {};
     }
 
+    const { email: authorEmail } = this.props.build.createdBy;
```
sj26: Woah, this syntax is interesting! I think I like it?

Is there any benefit over:

```js
const authorEmail = this.props.build.createdBy.email
```

ticky: Not really - initially I’d called the var `email` and I like the deconstruction syntax for that (vs. repeated “email” on the same line) but it’s a matter of taste, really

toolmantim:
Love this change so much!

I wonder if we should remove the "@" because outside of GitHub it could be a bit confusing?

> The GitHub account [ticky](https://github.com/ticky) could not be matched to any users in your organization.

Also the "Manage Connected Apps" might be a little confusing for people, because maybe they think that's an org specific thing? (not user specific). I know it won't directly be a "connect" action, and just link to the "Connected Apps" page, but perhaps it'd make the choice clearer to the people if we renamed it to be "Connect this GitHub Account" and "Not My Account" — and then in the future, that'd actually just kick off the GitHub auth process.

ticky:
The future in which this _could_ directly kick off the GitHub auth process is simply after buildkite/buildkite#2183 lands, at which point I’d also be updating this to use it as part of the decision making process for showing this prompt. I think landing this more textual update _first_ is a reasonable step!

I’ve updated this to remove the @ (I realise that the text above the prompt will, in real cases, show the same name as in the prompt, so my concern about the text not being very obvious are less realistic than I thought!) and update the button prompt text to “Connect GitHub Account”. (“Connect this GitHub Account” made the buttons overflow!)

screen shot 2017-06-14 at 10 47 56 am

PR #299, opened 2017-06-19
Pipeline paginator

ticky:
This is an extension of #291 which adds the “Show More” component to the Pipeline index.

I didn’t include it in #291 because it feels a bit _iffy_. Passing a component reference at render time feels really awkward and weird.

It also feels like, with the knowledge that we’re pondering design refreshes for these components, but those refreshes don’t appear to apply to the Pipelines page, this might need to be held back until we know what we want.

Or maybe I’m overthinking the whole thing 🤔


ticky:
Actually, never mind, I thought of a better way!

PR #300, opened 2017-06-21
Relay preload pipeline teams data

ticky:


app/lib/RelayPreloader.js
```diff
@@ -395,6 +395,46 @@ const QUERIES = {
       }
     }
   `,
+  "pipeline/settings/teams": Relay.QL`
+    query($pipeline: ID!) {
+      pipeline(slug: $pipeline) {
+        id
+        slug
+        name
+        organization {
```
keithpitt: I think you may also need to add `id` here for completeness (Relay always asks for an ID if there is one) - but it may have been cached so it didn't ask for it in _this_ request.

keithpitt: I think you may also need to add `id` here for completeness (Relay always asks for an ID if there is one) - but it may have been cached so it didn't ask for it in _this_ request.

PR #301, opened 2017-06-29
Fancy new tab control

ticky:


app/components/shared/TabControl/Tab.js
```diff
@@ -0,0 +1,47 @@
...
+  className: 'px3 pt2 pb1 border-bottom block text-decoration-none black',
+  activeClassName: ACTIVE_CLASS_NAME
+})`
+  border-bottom-width: 4px;
+  border-color: transparent;
+
+  &.${ACTIVE_CLASS_NAME} {
+    border-color: ${cssVariables['--lime']};
+  }
+`;
```
ticky: some comments I made on this stuff while I was querying Tim’s opinion on it:

> it requires enough stuff basscss itself can’t do alone that I’ve had to use it with styled-components, but also enough stuff styled-components and basscss can’t do alone that I’ve had to implement a small custom component

(this is why I’m not just using `border-lime` in `activeClassName`)

> it’s probably mostly good but it feels inelegant
> it’s mostly complicated by react-router not having good utilities for querying or reacting to state outside of `Router.Link`

app/components/team/Show.js
```diff
@@ -58,13 +67,60 @@ class TeamShow extends React.Component {
...
+          {this.renderTabs()}
+
+          {this.props.children}
         </div>
       </DocumentTitle>
     );
   }
 
+  renderTabs() {
+    const tabContent = permissions(this.props.team.permissions).collect(
```
keithpitt: Nice, I'm glad it worked out here as well :)

app/components/shared/TabControl/Tab.js
```diff
@@ -0,0 +1,47 @@
...
+
+const TabButton = styled(Link).attrs({
+  className: 'px3 pt2 pb1 border-bottom block text-decoration-none black',
+  activeClassName: ACTIVE_CLASS_NAME
+})`
+  border-bottom-width: 4px;
+  border-color: transparent;
+
+  &.${ACTIVE_CLASS_NAME} {
+    border-color: ${cssVariables['--lime']};
```
keithpitt: `cssVariables` very nice

ticky: Glad you like it! 😃

PR #302, opened 2017-07-05
Self-enable teams support

toolmantim:
- [x] Draw two circles
- [x] Front-end with placeholder copy


toolmantim:
@ticky I've removed Keith and my tasks from this for now, as you won't see this tab/page until the flag is turned on. Merge at will!

PR #304, opened 2017-07-07
Turn off teams button

keithpitt:
@toolmantim I kicked off this PR for you (to make sure that the the disabling worked in the backend). I wasn't sure what styling you had in mind so I just did something simple for now.

You can put it next to the "There are no teams in this organization" message if you want, but you'll just need to make sure that it doesn't show when you search for a team with no results (cause I think they're the same message)

screen shot 2017-07-07 at 2 10 27 pm


toolmantim:
Just updated it to this:

new-copy

toolmantim:
The intro/enable screen needs updating too to match, but that's a separate PR.

PR #305, opened 2017-07-07
Add tabs to "Member Edit"

keithpitt:
@ticky I used your `TabControl`, worked a treat!

screen shot 2017-07-07 at 3 54 57 pm

screen shot 2017-07-07 at 3 55 28 pm

screen shot 2017-07-07 at 3 55 41 pm

### DIFF
--- a/app/components/member/Edit/form.js
+++ b/app/components/member/Edit/form.js
@@ -36,8 +36,6 @@ class Form extends React.PureComponent {
   }
 
   render() {
-    let content;
-
     if (!this.props.organizationMember.permissions.organizationMemberUpdate.allowed) {
       return (
         <Panel>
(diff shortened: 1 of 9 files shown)


app/components/member/Edit/remove.js
```diff
@@ -41,32 +41,25 @@ class Remove extends React.PureComponent {
...
+        </Panel.Section>
+      </Panel>
     );
   }
 
   handleRemoveClick = () => {
-    const message = this.props.isSelf
-      ? 'Removing yourself will immediately revoke your access to this organization.'
-      : 'Removing this user will immediately revoke their access to this organization.';
+    const message = 'Removing this user will immediately revoke their access to this organization.';
```
ticky: Is there a reason you removed the handling of removing oneself?

app/components/member/Edit/form.js
```diff
@@ -12,8 +12,8 @@ import OrganizationMemberUpdateMutation from '../../../mutations/OrganizationMem
 
 import OrganizationMemberRoleConstants from '../../../constants/OrganizationMemberRoleConstants';
 
-class Role extends React.PureComponent {
-  static displayName = "Member.Edit.Role";
+class Form extends React.PureComponent {
+  static displayName = "Member.Edit.Form";
```
ticky: I’m wondering if the `displayName = blah` pattern is the wrong approach here. Given the Relay compiler seems to insist component names are their file name. It’s real irritating :(

keithpitt: Yeah totally. I think Relay Modern wants them to be named the right thing, so in this case, the component would be called: `MemberEditForm`. Maybe if we did a sweep and updated names of all the things that'd be a good idea?

app/components/member/Edit/remove.js
keithpitt: As I was going through the pages of all the users, it'd say one thing, but then when you look at yourself, all the words changed, and my brain had to re-parse what the button was saying. My feels were we didn't gain much for the overhead - and it's probably to be consistent (also makes for easier unit testing since it's the one button we test for, not 2)

PR #306, opened 2017-07-09
Respect `pipelineCreate` permission on the organisation show page

keithpitt:
If you don't have permission, we shouldn't show the button!


keithpitt:
There's nothing too controversial in this change so I'm just gonna merge it in. It's servicing a pretty rare edge case.

PR #309, opened 2017-07-10
Teams form tweaks

keithpitt:
Things I've done in this PR:

- [x] Break teams form into sections
- [x] Add "Required" to the Team Name input
- [x] Add UUID to the "Edit" version of the form
- [x] Only focus on the "Name" when creating a team (don't autofocus on edit)
- [x] Added "Recommended" to the "Visibility" option
- [x] Don't redirect away from the "Settings" tab when saving a team - just flash.

Here's a screen grab of the "New" page:

screen shot 2017-07-10 at 2 46 22 pm

And the "Edit" page:

screen shot 2017-07-10 at 2 46 43 pm

### DIFF
--- a/app/components/team/Edit.js
+++ b/app/components/team/Edit.js
@@ -2,6 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Relay from 'react-relay/classic';
 
+import Panel from '../shared/Panel';
+
 import TeamForm from './Form';
 import TeamDelete from './TeamDelete';
 
@@ -45,7 +47,6 @@ class TeamEdit extends React.Component {
       <div>
         <form onSubmit={this.handleFormSubmit}>
           <TeamForm
-            uuid={this.props.team.uuid}
             onChange={this.handleFormChange}
             errors={this.state.errors}
             name={this.state.name}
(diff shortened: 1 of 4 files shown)


toolmantim:
Nice! Thanks!

Two things:

* Can we just put the UUID part in it's own panel below the form?
* I think we need to rename that "Default" label to something! How about we remove the section header for that bit, and change the checkbox to label and description to:

```
- [ ] Default team
      Automatically add new users to this team when they join the organization.
```

That way there's a clear "When this option is ticked, the 'Default' label is shown"

keithpitt:
@toolmantim yeah good ideas! I've made those changes and it looks like this now:

screen shot 2017-07-10 at 3 50 14 pm

app/components/team/Edit.js
```diff
@@ -45,24 +46,26 @@ class TeamEdit extends React.Component {
...
+            isDefaultTeam={this.state.isDefaultTeam}
+            saving={this.state.saving}
+            button={this.state.saving ? "Saving team…" : "Save Team"}
+          />
         </form>
 
+        <Panel className="mt3">
+          <Panel.Header>REST API Integration</Panel.Header>
+          <Panel.Section>
+            <p>You can use this UUID to reference this team when using the <a href="/docs/rest-api/pipelines#create-a-pipeline" className="blue hover-navy">Create a Pipeline</a> REST API.</p>
```
toolmantim: My only change. "Create Pipeline REST API" rather than "Create a Pipeline" (even though I know that the head in the doc says that). I think in this case it just reads better, and we do that elsewhere sometimes too)

keithpitt: Yah cool, made that change as well. Will 🚀 on 🥗

PR #311, opened 2017-07-10
Show labels on user invite team selection

toolmantim:
This updates the team selector on the user invite screen. It now shows the full team description, as well as the default/private/public labels.

invite

Next step is probably to use this as our standard "Team selector" UI. And wherever else we were using the "selected green border" type UI!


toolmantim:
Gonna merge this, so @sj26 can use it in other stuff he's doing

PR #315, opened 2017-07-10
UI for Audit Events

ticky:


app/components/organization/SettingsMenu.js
```diff
@@ -101,6 +101,18 @@ class SettingsMenu extends React.Component {
         )
       },
       {
+        allowed: "teamAdmin",
```
keithpitt: Hrm... I don't think `teamAdmin` is appropriate for this one (since that'll allow non-org admins to view logs). We'll need to add a new permission, but we can do that in another PR.

So for now, let's use `agentTokenView`

keithpitt: `agentTokenView` is scoped to organization administrators

sj26: @ticky we have [a `see_account_audit_events` permission](https://github.com/buildkite/buildkite/blob/audit-ui-permission/app/models/permission.rb#L190-L192), maybe we need to [expose it](https://github.com/buildkite/buildkite/compare/master...audit-ui-permission) to use it here?

app/components/audit_log/row.js
```diff
@@ -0,0 +1,279 @@
...
+      .split(/\s+/);
+
+    const eventVerb = eventTypeSplit.pop().toLowerCase();
+
+    const eventType = eventTypeSplit.join(' ');
+
+    let subjectName = subject.name;
+
+    if (eventTypeName === 'AuditOrganizationCreatedEvent') {
+      subjectName = `${subjectName} 🎉`;
```
sj26: Cute!

PR #316, opened 2017-07-11
Use <Badge outlined> for team labels

toolmantim:
Also used in an upcoming pull request…


ticky:
I wish we didn’t have that lint rule where boolean things need to be passed an actual boolean so `<Badge outlined />` was allowed :P

toolmantim:
Yeah! Let's change it?

ticky:
I suppose the trouble comes when you have `false` things which have to be fully qualified and it looks a bit weird? IDK.

toolmantim:
Let’s ditch it the rule!

PR #318, opened 2017-07-11
Tab layout tweaks

toolmantim:
This makes a bunch of styling updates to the `<TabControl>`:

* Coloured/active bottom border goes over the grey line
* Hover styles
* Padding/margin tweaks

I'm still not 100% happy with bottom margin on it, but it seems like works best with panel bodies that also have buttons right below it.

Before:

before

After:

after


ticky:
Wait, I’m confused, isn’t this how it looked before your designs got iterated on???

toolmantim:
Yeah, I did some more designing in browser and settled up these updates instead.

PR #334, opened 2017-07-12
Add overflow scroll to autocomplete dialog

toolmantim:
If you have a large number of users, the dialog is a bit unwieldy. And it's height suggests almost helps to reinforce the misconception that it might contain all users? (another issue)

team-selector-before

This fixes the height to half the viewport height, and adds a overflow scroll:

teams-after

I wanted to treat the empty case differently… say, for a small account, or if there are "no more users to add", but unfortunately I couldn't wrangle the components, so will leave that for another PR.

boo-1

boo-2

### DIFF
--- a/app/components/shared/Autocomplete/Dialog/index.js
+++ b/app/components/shared/Autocomplete/Dialog/index.js
@@ -100,7 +100,7 @@ class AutocompleteDialog extends React.PureComponent {
     }
 
     return (
-      <div className="block" style={{ width: "100%", minHeight: 300 }}>
+      <div className="block" style={{ width: "100%", minHeight: 300, maxHeight: '50vh', overflow: 'auto' }}>
         <ul className="list-reset m0 p0">
           {suggestions}
         </ul>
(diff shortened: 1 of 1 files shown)


app/components/shared/Autocomplete/Dialog/index.js
```diff
@@ -100,7 +100,7 @@ class AutocompleteDialog extends React.PureComponent {
     }
 
     return (
-      <div className="block" style={{ width: "100%", minHeight: 300 }}>
+      <div className="block" style={{ width: "100%", maxHeight: 250, height: '50vh', overflow: 'auto' }}>
```
ticky: this isn’t going to be good on mobile 🙈

toolmantim: I tested it, it's not too bad! Any suggestions for making it less crazy on desktop, when you have a larger screen and dozens of users in the org? (It's tedious to set that data up for dev, but it appears we never designed for those cases)

There's actually lots of other teams stuff that's more broken on mobile too! But I didn't get time to fix much of it 😞

PR #335, opened 2017-07-12
Relay data fixes for Chooser dialogs

keithpitt:


app/components/member/Teams/chooser.js
```diff
@@ -98,25 +100,23 @@ class Chooser extends React.Component {
...
-          showingDialog: true
-        }, () => {
-          this._autoCompletor.focus();
-        });
+        this.setState({ loading: false });
       }
     });
+
+    // Now start showing the dialog, and when it's open, autofocus the first
+    // result.
```
ticky: “the first result” doesn’t seem to be what this does?

app/components/member/Teams/chooser.js
```diff
@@ -168,7 +168,7 @@ class Chooser extends React.Component {
 
 export default Relay.createContainer(Chooser, {
   initialVariables: {
-    showingDialog: false,
+    isMounted: false,
```
ticky: Why did you change this to `isMounted`? This _doesn’t_ occur when the dialog is mounted

app/components/member/Teams/chooser.js
keithpitt: Lol, this is a bad comment - I meant to say focus the search input. Will fix my comments!

app/components/member/Teams/chooser.js
keithpitt: Oh you're totally right, I was rushing through some last minute changes and this is poorly named. I'll change them all back to `showingDialog` which is actually what it _is_

PR #337, opened 2017-07-12
Tweak Teams pager header icon size

toolmantim:
Whilst doing the screenshots for the announcement, the icon size and margin for the Teams section was bugging me. So this PR reduces the icon size in the header, and tweaks the margins.

Before:

before-1

before-2

After:

after-1

after-2

### DIFF
--- a/app/components/member/Index.js
+++ b/app/components/member/Index.js
@@ -72,7 +72,6 @@ class MemberIndex extends React.PureComponent {
             <PageHeader.Icon>
               <Icon
                 icon="users"
-                className="align-middle mr2"
                 style={{ width: 40, height: 40 }}
               />
             </PageHeader.Icon>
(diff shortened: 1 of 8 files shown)


ticky:
Hmm, this means the layouts of the header differ between each page. Is this desired?

toolmantim:
@ticky did I mess up the calculations? The icon should be different, but everything should still add up! Feel free to fix any typos you found.

The idea of moving them all to the shared component was to make sure they all the same layout! It's just some icons need optical relative size adjusting, and I couldn't do it in the icon file itself.

ticky:
Users’ `section.flex.flex-wrap.items-top.mb2 > div.flex-none.mb2:first-child` is 53px wide. Same element on Teams is 50px wide.

If you wanna work on icons which can adjust to compensate for size let’s definitely make that happen!

toolmantim:
I wanted to just change the source icon itself, so it doesn't take up the full 20px view box (they should be optically adjusted as a complete set), but didn't have time. Didn't mean to break it tho.

I want to lay all the icons out ala React Storybook.

PR #339, opened 2017-07-12
Audit log production readiness

ticky:


keithpitt:
Ah, I just had another look and I think you need to do the `isMounted` tricky so the data only loads in once the component has mounted (similar to the Users list)

ticky:
Right you are! Good catch 😊

PR #340, opened 2017-07-13
Further Relay fixes for Chooser dialogs

keithpitt:
Updates based on your feedback @ticky!

### DIFF
--- a/app/components/pipeline/teams/Index/chooser.js
+++ b/app/components/pipeline/teams/Index/chooser.js
@@ -23,9 +23,6 @@ class Chooser extends React.Component {
         teams: PropTypes.shape({
           edges: PropTypes.array.isRequired
         })
-      }),
-      teams: PropTypes.shape({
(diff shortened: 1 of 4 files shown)


app/components/pipeline/teams/Index/chooser.js
```diff
@@ -192,16 +189,6 @@ export default Relay.createContainer(Chooser, {
             }
           }
         }
-        teams(first: 500) @include (if: $isMounted) {
```
keithpitt: I think this data may have been left over from a time where we didn't have your selector technology @ticky - can you think of why we need it?

ticky: I suspect you’re right! I can’t see it being used :)

keithpitt: Awesome :) I think back in the day we got all the teams for the pipeline, and all the organisation pipelines, and just did a diff in JS :)

PR #342, opened 2017-07-17
Clean up the audit log reveal chevron

toolmantim:
This extracts out the copy-pasta rotatable chevron code from Create Build and Audit Logs into a shared icon component, and in the process cleans up the Audit Log chevron sizing.

Before:

before

After:

after

### DIFF
--- a/app/components/shared/Icon/RevealableDownChevron.js
+++ b/app/components/shared/Icon/RevealableDownChevron.js
@@ -22,7 +22,8 @@ export default class RevealableDownChevron extends React.PureComponent {
     revealed: PropTypes.bool,
     size: PropTypes.number,
     strokeWidth: PropTypes.number,
-    className: PropTypes.string
+    className: PropTypes.string,
+    style: PropTypes.object
   };
 
   static defaultProps = {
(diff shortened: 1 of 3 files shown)


app/components/shared/Icon/RevealableDownChevron.js
```diff
@@ -0,0 +1,47 @@
...
+  trasform-origin: center 0;
+  transition: transform 300ms;
+  will-change: transform;
+  width: ${(props) => props.size}px;
+  height: ${(props) => props.size}px;
+
+  polyline {
+    stroke-width: ${(props) => props.strokeWidth}px;
+  }
+`;
```
ticky: I feel like you could just export this `RotatableIcon` itself and have the same functionality, with `defaultProps` specified. Bonus points for attaching `propTypes`!

toolmantim: I tried! But couldn't get it work with customizable className for color setting. I tried 3 variations! And with propTypes.

ticky: That’s weird, because we’re not actually overriding it anywhere?

ticky:
Yeah, I saw your comment about wanting to actually measure the contents. Definitely something we _should_ do sometime :P

PR #346, opened 2017-07-20
Audit descriptions

sj26:
Some attempts at more event types in descriptions:

screen shot 2017-07-20 at 5 41 35 pm

I was playing around with some styles, too, and omitting fields without data. React won't let me return a dt/dd without a parent so I had to make multiple dls :-(((

Also "an user", hmm. Might need a special case.

### DIFF
--- a/app/components/audit_log/Drawer/subjectSection.js
+++ b/app/components/audit_log/Drawer/subjectSection.js
@@ -26,55 +26,51 @@ class AuditLogSubjectSection extends React.PureComponent {
         <SectionHeading className="m0 mb2">
           Subject
         </SectionHeading>
-        {this.renderSubjectType(subject)}
-        {this.renderSubjectName(subject)}
-        {this.renderSubjectUuid(subject)}
+        <dl className="m0">
+          {this.renderSubjectType(subject)}
+          {this.renderSubjectName(subject)}
+          {this.renderSubjectUuid(subject)}
+        </dl>
       </Section>
     );
   }
 
   renderSubjectType({type}) {
-    return (
-      <dl className="m0">
-        <dt className="mt1 dark-gray">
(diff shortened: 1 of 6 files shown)


app/components/audit_log/row.js
```diff
@@ -204,6 +246,17 @@ export default Relay.createContainer(AuditLogRow, {
...
+            ...on Team {
+              name
+            }
+            ...on TeamMember {
+              team { name }
+              user { name }
+            }
+            ...on TeamPipeline {
+              team { name }
+              pipeline { name }
```
ticky: I know it doesn’t affect much but for the sake of future diffs can we keep each query property on a separate line? 😄

sj26: Considering I'm a stickler for hanging commas for the same reason, I endorse this product and/or service. :+1:

app/components/audit_log/Drawer/subjectSection.js
```diff
@@ -26,28 +26,56 @@ class AuditLogSubjectSection extends React.PureComponent {
...
-            {subject.type}
-          </dd>
-          <dt className="mt1 dark-gray">
             Subject Name
           </dt>
           <dd className="ml0">
-            {subject.name}
+            {name}
           </dd>
+        </dl>
```
sj26: Do we need key props if the collection will never change? That's just to optimise when rows are likely to change around other rows, right?

sj26: And of course I can use arrays 🤦🏻‍♂️

sj26:
Looks like this, now:

screen shot 2017-07-21 at 4 24 51 pm

ticky:
Heck yeah ✌🏼

PR #348, opened 2017-07-26
Add a basic credit card input type

ticky:
- [x] Automatic credit card formatting for all supported types
- [x] Don’t push the cursor to the end of the field when typing in the middle
- [x] Use `credit-card-type` output to highlight card type in main form
- [x] Update “CVC” label and max length to match input credit card types
- [x] Improve validation and mobile input (Read: show numeric keypad)
- [x] Add specs for the input behaviour
- [ ] Check it works for AutoFill
  - [x] 1Password
  - [ ] LastPass
  - [ ] Safari
  - [x] Chrome
  - [ ] Firefox
  - [ ] Microsoft Edge


toolmantim:
What about "Make sure it works with 1Password, LastPass, Safari and Chrome?"

ticky:
Oh, Tim, like that’s even optional 😉

PR #354, opened 2017-08-11
Fix typo (CloudFormtion => CloudFormation)

bendrucker:


ticky:
Nice catch, @bendrucker :smile:

PR #356, opened 2017-08-15
Form element overhaul

ticky:
This updates all our text form components to share a single set of higher-order components. `collapsibleFormComponent` and `labelledFormComponent` accept form elements and decorate them with labels, error handling, and help text support. Both are fully tested.

Into these, our forms pass either normal form elements (`input`/`textarea`) or components which further wrap these primitives and add behaviours, like `CreditCardInput` (formerly `FormCreditCardField` or `AutosizingTextarea`. These both add custom behaviour to the primitives, and conform to the subset of the DOM API which our form elements expect to be available. Namely, a `value` getter, and a `focus` method.

I was going to add an emoji input decorator based on these changes, but I’ve since discovered it’ll require some back-end refactors so I’ll separate that out.

- [x] Consolidate `FormInputHelp` & `FormInputErrors`
- [x] Simplify and centralise shared input element behaviours using a Higher-Order Component (read: component factory, lol)
- [x] Add specs
- [x] ~(STRETCH GOAL) Emoji picker 🌈:sparkles:🤔~ (moving this to a new PR as it requires further work)


ticky:
## Fields which accept emoji

- Project
  - name
  - description

- PipelineSchedule
  - label (AKA description)

- Team
  - name
  - description

- Build
  - message

- Step
  - name

- Job
  - name
  - command
  - prompt
  - submit
  - field.hint
  - field.option.label
  - field.label
  - log_deleted_message

- Agent
  - name

keithpitt:
I didn't end up looking much of the internal changes, I mostly care about the API for the components, and that seems mostly untouched. So I'm totally cool with the cleanup!

ticky:
The API for components was fundamentally brought more in line with native React DOM elements, which IMO will make things more consistent in the future! No more `value` which is actually `defaultValue` lol

Did you give this a play? I’ve given all the places I can see using this a once-over, but I suspect you might know more?

toolmantim:
Nice cleanups @ticky! 🌈

ticky:
Thanks @toolmantim 😄

PR #357, opened 2017-08-18
Fix showing seconds in FriendlyTime when capitalizing words

keithpitt:
@ticky I wasn't 100% sure on how to fix this. I found a bug where `seconds` wouldn't have an affect if `capitalised` was being set to true.

### DIFF
--- a/app/components/shared/FriendlyTime.js
+++ b/app/components/shared/FriendlyTime.js
@@ -69,7 +69,7 @@ export default class FriendlyTime extends React.PureComponent {
       <time
         className={this.props.className}
         dateTime={this.props.value}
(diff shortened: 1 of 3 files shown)


app/components/shared/FriendlyTime.js
```diff
@@ -69,7 +69,7 @@ export default class FriendlyTime extends React.PureComponent {
       <time
         className={this.props.className}
         dateTime={this.props.value}
-        title={getDateString(this.props.value)}
+        title={getDateString(this.props.value, true)}
```
keithpitt: I decided to always show seconds in the tooltip here.

ticky:
Oh, good catch!

PR #358, opened 2017-08-23
Move webpack dist into docker build

lox:
This is potentially controversial. 

The thinking behind this is to roll a docker container with the production js/css in it, such that the buildkite/buildkite docker stuff can depend on something that is known good and working. To do this, it runs webpack in the Docker build process and puts the `dist/` dir in the container and serves it with a plain 'ol http server rather than webpack-dev-server.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,11 @@ FROM node:6
 
 EXPOSE 4890
 
-RUN echo "--- :watchman: Installing watchman" \
-    && cd /tmp \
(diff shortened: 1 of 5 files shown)


ticky:
Won’t this change stop hot reloading from working?

lox:
@ticky yup, it's not designed for live-dev, just for "known working, instant on" frontend container. We can add a live-reloading, code mounting, webpacking on file changing container as an additional thing (was my thinking).

PR #359, opened 2017-08-23
Switch to direct mutations for deleting team members

keithpitt:
This fixes various issues with Relay being a bit silly when it tries to reload connections.


keithpitt:
Sorry, I should have marked this as WIP - cause it wasn't quite done yet. I actually broke a few things (like numbers in badges updating with this change). We may need to review it!

toolmantim:
@keithpitt can do

PR #362, opened 2017-08-26
ES6ify pipeline header components

ticky:
This was part of a scheme to try to move these components’ data requirements to GraphQL.

This part ended up taking a bit longer than I’d hoped, but I guess it might still be a step in the right direction.

Components were converted using [Decaffeinate](http://decaffeinate-project.org), and [react-codemod](https://github.com/reactjs/react-codemod), with guest star [Prettier](https://prettier.io), and some hand-massaging to add PropTypes and make it pass our rigorous ESLint suite.

`BootstrapTooltipMixin` also made the jump despite being _quite_ legacy because it was required by one of the components I needed to move, and it otherwise isn’t loaded in time for the ES6 code to use it.


toolmantim:
Did I say this was amazing? I should have, because it is.

PR #363, opened 2017-08-28
Change unknown email prompt to something clearer

lox:
I've clicked this button many times incorrectly, and have seen lots of other people do it. I don't love this copy, but it's at least clearer. 

Screenshot


keithpitt:
I think it's a rad change!

PR #365, opened 2017-09-05
Add @dhau/relay-compiler-webpack-plugin

lox:
Looks like @dhau has a fork of the relay-compiler-webpack-plugin that supports relay-compiler 1.1.0 and also doesn't require watchman optionally. This PR brings in the namespaced package (upstream has stopped responding).


lox:
Any ideas what caused all of these failures @ticky?

ticky:
Well, there’re [some lint errors in the webpack config](https://buildkite.com/buildkite/frontend/builds/2276#f89ab731-3ba1-4a34-bb78-59bb884c5721). I’m not sure what changed about Badge, though.

Also looks like the new plugin is expecting a `relay` executable in the PATH, which we don’t appear to be installing?

lox:
Totally missed those linting errors under the warnings!

lox:
I assume the extra output in the json profile analysis is a relay-compiler-webpack-plugin error, have filed an issue: https://github.com/danielholmes/relay-compiler-webpack-plugin/issues/5

PR #366, opened 2017-09-06
Fix agent searching for upper case tag names/values

toolmantim:
If you start an agent with tag `tag=Value` and then try to search for it in the agents UI, you can only match it with `tag=value` — if you type `tag=Value` then it shows "0 matching agents"

tag-search

This fixes the problem by making, which makes the search case sensitive in the same way the step agent targeting rules are.

tag-searching-fixed

Search is still slightly different to agent targeting, in that partial matches are allowed, and you can search for an agent name case insensitively. But I think this is an improvement.

The alternative is to make tag searching completely case insensitive. But I'm afraid that might lead people to think that agent targeting rules are the same?

### DIFF
--- a/app/components/agent/Index/agents.js
+++ b/app/components/agent/Index/agents.js
@@ -126,10 +126,12 @@ class Agents extends React.PureComponent {
     // Parse the search. We only want to do this once, outside of the agent
     // loop, in case they have lots of agents
     const queries = localSearchQuery.map((string) => {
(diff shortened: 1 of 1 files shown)


toolmantim:
I'm going to merge this so at least `Hyperion=True` works, but if anyone thinks that it should be completely case insensitive then I'm happy to revisit and change it again!

lox:
Awesome! Nice one!!

PR #367, opened 2017-09-07
Wanted an apostrophe not a quote

lox:
Fixing a typo from my previous PR.


toolmantim:
UTF-8 — “Just Use It”™

lox:
I did, I think I changed it because of linting errors.

toolmantim:
We should probably change that linting rule!

ticky:
Also it’s JSX so you don’t need to quote these things at all!

PR #369, opened 2017-09-13
Add :flowtype: Flow type checking

ticky:
Flow is Facebook’s type checker. Designed and intended for use with React and ES2015+ projects, it’s pretty well-suited to what we’re building.

Types are often able to be inferred, but in many places annotating the types of objects and attributes will be necessary. I’ve made a decent start at this already as part of this pull request!

Backwards-compatibility with code without type annotations is preserved by allowing files to not be checked by not including an `@flow` comment at the start of the file.

I started looking at Flow in particular because of silly little bugs like this one: <https://github.com/buildkite/frontend/commit/91c8b89198e26d908d2dbab3800f92a9cb98d91e#diff-395b062ee1f62973de3415c62c9e393aL71>. Flow would’ve handily detected this error before we shipped the code, because it knows what methods are available at run-time.

It’s also useful for those times we step outside React’s model. There are several places where we set instance variables on React components to DOM references, as pushing them into the component’s state is undesirable. In these situations, we can (and _must_!) now explicitly declare the loose variables we’re setting in addition to the Props and State. This change is also visible in [Dropdown.js](https://github.com/buildkite/frontend/pull/369/files#diff-395b062ee1f62973de3415c62c9e393aR52).

There are a few main caveats:
- **We must currently declare type information for component props twice.**
   Once in the format Flow understands, and once in the way React understands. The Flow type information may well be more useful to us as time goes on, so it may in fact be better for us to standardise on Flow type declarations instead of React type declarations.
- **Flow does not understand the object bind syntax I use for Metrick.**
   This means we either need to not use Flow with these files (undesirable, obviously!) or roll back dependence on the `stage-0` Babel plugin (which means a return to magic numbers). Unfortunately Flow supplies its own parser (and is written in another language altogether), and can’t be plugged into Babel like ESLint can, so there’s not currently a workaround. Flow’s issue about support for these stage-0 properties is at facebook/flow#1545.
- **Flow type information will only be as useful as how much we put in.**
   This would be true of any type system, but given how large a JavaScript codebase we have, and given the disconnect between it and the CoffeeScript still lurking in our backend repo, holes will remain in this approach for some time to come.
- **Flow doesn’t magically know anything about our GraphQL data.**
  This seems like an obvious omission! There [might be ways around it](https://github.com/mixcloud/relay-flow-types), but obviously it’d be great if we could have this be automated somehow.

I think ultimately that Flow will be a relatively unobtrusive inclusion which will help us write more reliable JavaScript and ultimately pay dividends the more we make use of it. The _other_ bonus is that if we do eventually decide we don’t like it much, rolling it back is very simple!


ticky:
I’m liking this so far, though it feels like our code might hit too many caveats. Namely:
- Flow’s parser seems to be unable to parse `export FooBar from './foo-bar'`. At all. It just complains.
- Typed objects acquired from an untyped file _aren’t_ type checked.
- The React PropTypes linter doesn’t yet understand Flow’s Prop type definitions, so you must include both, and make sure they match unless _everything_ which uses your class is Flow typed(?). Is having both desirable or even reasonable? Not sure.

So more or less it’ll require a bunch of updates in order to have it pay dividends. I’m still trying to figure out if there are codemods out there to help automate the transition!

ticky:
Also, the inclusion of the `flow-typed` directory _appears_ to be best practice. I’m honestly surprised it doesn’t use some sort of lockfile instead? But I guess that’d be implausible for the type definitions it automatically generates. Weird stuff.

toolmantim:
This is cool!

So this gives us this eh: https://buildkite.com/buildkite/frontend/builds/2286#da36fa08-f3a0-4959-94ef-856af7c5d0a2

flow-type

Also, jeez we have a lot of noisy groups in that build output.

ticky:
>So this gives us this eh: https://buildkite.com/buildkite/frontend/builds/2286#da36fa08-f3a0-4959-94ef-856af7c5d0a2
>
>flow-type

Yeah! Flow will check the types on everything we’ve explicitly enabled it on, which means it won’t give us benefit on any file which isn’t annotated with type information, and doesn’t begin with an `@flow` comment.

The benefit going forward is that silly things like “this was misspelled” or “I renamed a thing in one place but not another” can be caught in static analysis, where it can’t currently. It also “encourages” us to be more careful about how we deal with Javascript’s types at large, as it’ll yell at you if something’s being used weirdly.

There’s also an ESLint plugin for Flow syntax, which I’ve enabled with the “recommended” settings. I think we can be much more rigorous with it, but I think this is good for getting settled in with!

>Also, jeez we have a lot of noisy groups in that build output.

We sure do! haha. I do wish some of the docker stuff were collapsed into a smaller thing, but I guess then you’d lose some of the breakdown stuff we have? Do we need multi-level log folding? 😅

I’ve been mulling over whether to graduate this from being just an experiment - adding type annotations to our codebase will be a long-term change for a relatively small set of benefits. Additionally, I’m nervous given we’re already strung between about three “generations” of code (CoffeeScript, JS+jQuery.ajax, React+GraphQL) and I don’t want to introduce much more complexity. It **also** requires us to maintain the `flow-typed` directory (with the pretty easy to use `flow-typed` utility, thankfully, but still).

The more I consider it, though, the more it feels like this is a _relatively_ unobtrusive way to start doing this, and the benefits down the line in terms of confidence in our code are pretty good. In particular, this feels like a slightly easier way to improve our confidence in a frontend change given our relatively low test coverage - if it passes some basic static analysis we maybe haven’t broken it, haha.

ticky:
Further caveats: my function bind syntax hacks are not supported by Flow without workarounds (see facebook/flow#1545)

ticky:
In the build for eb533d1 (https://buildkite.com/buildkite/frontend/builds/2292#43e97944-4a85-466b-bfe4-3ce195a2d7f4), you can see failures from both linting and output from Flow’s static analysis. It’s pretty sweet, I think!

screen shot 2017-09-14 at 11 40 10 am

ticky:
facebook/flow#1545 may be the biggest blocker here… it means we simply can’t properly use Flow with any files which use the function bind syntax.

keithpitt:
I'm 100% on board with this. What ever you need to do to change the code to make this work, I fully support.

keithpitt:
Just did a quick squiz of the changeset, and it seems we have to vendor a ton of flow type definitions? Is there another way to do that which doesn't bloat the source tree?

(there probably isn't, just thought I'd ask...)

ticky:
> we have to vendor a ton of flow type definitions? 

Unfortunately this seems to be a fact of life with Flow right now. `flow-typed` doesn’t appear to guarantee it will ship type definitions for any arbitrary version of a dependency indefinitely, so vendoring seems to be necessary.

We could instead have our Docker image pull in the `flow-typed` definitions, but I don’t know how reliable that’s likely to be going forward, and it means we can’t then update definitions ourselves very easily.

toolmantim:
Could we put them on s3, and be able to push/pull from CI and/or locally?

toolmantim:
(not that I think that s3 is necessarily a good idea, or that putting it in source is a prob)

keithpitt:
> Unfortunately this seems to be a fact of life with Flow right now. flow-typed doesn’t appear to guarantee it will ship type definitions for any arbitrary version of a dependency indefinitely, so vendoring seems to be necessary.

Ah right, I'm sure it's totally fine while we're still finding out feet. When the world changes so we don't need to do that - we can just `rm` it and have a party 🚀

ticky:
> Could we put them on s3, and be able to push/pull from CI and/or locally?

I don’t think s3 would make this any less complicated - in fact I’d suggest it’d make versioning and deploys much _more_ complicated than just using them checked in.

ticky:
I’ve added a bunch of type annotations to this, and I’ve run the specs against it from backend, I think we’re good to go.

toolmantim:
Rad!

It looks like we should be relying on Flow to do our `propType` checking now, rather than defining it almost twice for each component?

ticky:
> It looks like we should be relying on Flow to do our `propType` checking now, rather than defining it almost twice for each component?

That is likely a future path, but Flow’s types will only be statically analysed when they are used from a `@flow` annotated file. Until we have much, much more of the codebase updated, we’ll likely need to rely on both!

toolmantim:
Okie dokie 😭

ticky:
I’m going to merge this because I believe it’ll be worthwhile and the longer it stays out of master the less likely it is we’ll make it happen :)

toolmantim:
Do it!

benhinchley:
Another option you have for generating type definitions for your graphql schema is [apollo-codegen](https://github.com/apollographql/apollo-codegen).

PR #371, opened 2017-09-14
A potential approach to Flow+Babel editor integration

ticky:
Having dug into the issues in facebook/flow#1545, this is the closest I’ve gotten to making it work. It’s more or less there, with output working okay in terminal and also showing up in SublimeLinter, however, the positions of the highlights in SublimeLinter are offset due to the fact that Babel reformats our code slightly when it transforms the bind syntax.

Unless there’s something I’m missing, this is already doing the minimum we can ask Babel to do.

Done as a separate PR from #369 so I can test it in CI.


plasticine:
I think this got done in the end? Could we close this?

ticky:
Yeah this was an attempt to improve the editor integration but I don’t think it matters any more!

PR #374, opened 2017-09-21
Restore bundle analysis functionality

ticky:
Working around danielholmes/relay-compiler-webpack-plugin#5 by avoiding stdout for JSON generation, using stats-webpack-plugin.

Related: #365.


ticky:
#yolomerging because this doesn’t change much

PR #384, opened 2017-11-10
Add text-decoration-none to pipeline steps.

KevinGrandon:
There's currently a few bugs with the text-decoration on pipeline steps. Disabling text-decoration seems to fix it.

Bug 1 - Drag and dropping emoji from a pipeline steps shows an underline.
Bug 2 - Clicking on text pipeline steps seems to randomly show an underline.

Example error:
image

### DIFF
--- a/app/components/build/Header/pipeline.js
+++ b/app/components/build/Header/pipeline.js
@@ -95,7 +95,7 @@ const BuildHeaderPipelineComponent = createReactClass({ // eslint-disable-line r
         <a
           key={`${this.props.build.id}-more-jobs`}
           href={this.props.build.path}
-          className="build-pipeline-job truncate align-middle"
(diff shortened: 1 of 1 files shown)


balthazar:
This is blocking us to completely switch to buildkit at our company, I'm so happy someone got around that!

toolmantim:
Thanks for the PR! The underline, it hurts my eyes! And so does the hover color.

So it turns out it was lacking a `:focus` style, and it wasn't this part of the code. But the fix is live 🙌🏼

Thanks @KevinGrandon — when more of the build page UI is in this repo, you'll be able to get some PRs merged. Feel free to keep giving it a shot though, and don't be afraid to ping us in Slack before you do so if you want to check it's the right place to be looking in the code.

PR #385, opened 2017-11-16
Everyone can see build badge settings

toolmantim:
This allows everyone to see the Pipeline Settings and Build Badges menu items.

### DIFF
--- a/app/components/pipeline/Header/index.js
+++ b/app/components/pipeline/Header/index.js
@@ -158,7 +158,7 @@ class Header extends React.Component<Props, State> {
         })
       },
       {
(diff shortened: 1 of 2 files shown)


app/components/pipeline/Header/index.js
```diff
@@ -158,7 +158,7 @@ class Header extends React.Component<Props, State> {
         })
       },
       {
-        allowed: "pipelineUpdate",
+        always: true,
```
ticky: Does this look okay on mobile with one option in the menu? 🤔

toolmantim: Yeah, okay "enough"

PR #387, opened 2017-11-17
Add a detailed job history for Agents

ticky:
This adds a detailed history of jobs run on each Agent. Each job is shown, along with the time it was assigned to the agent.

screen shot 2017-11-20 at 12 05 08 pm

It has full pagination support:

screen shot 2017-11-20 at 12 05 19 pm

The zero-job case is handled as shown (though in the zero-job case there is no link to this view, so it will only be shown to those savvy enough to go exploring):

screen shot 2017-11-20 at 12 05 27 pm

Clicking anywhere in the panel will take the user to the build page the job is from, with the job log opened.

The existing “Recent Jobs” section has been updated to match, and include two links to the new page (the “and n more” text, and the title which is included for when the overflow text isn’t shown):

screen shot 2017-11-20 at 12 04 58 pm

fixes buildkite/feedback#286

### DIFF
--- a/app/components/agent/Jobs.js
+++ b/app/components/agent/Jobs.js
@@ -1,7 +1,7 @@
 // @flow
+/* eslint-disable react/prop-types */
 
 import React from 'react';
-import PropTypes from 'prop-types';
 import Relay from 'react-relay/classic';
 import { Link } from 'react-router';
 import DocumentTitle from 'react-document-title';
@@ -40,7 +40,7 @@ type Props = {
 
 type State = {
   loading: boolean
-}
+};
 
 class AgentJobs extends React.PureComponent<Props, State> {
   state = {
--- a/app/components/agent/Jobs.js
+++ b/app/components/agent/Jobs.js
@@ -1,0 +1,138 @@
+// @flow
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import Relay from 'react-relay/classic';
+import { Link } from 'react-router';
+import DocumentTitle from 'react-document-title';
+
+import FriendlyTime from '../shared/FriendlyTime';
+import JobLink from '../shared/JobLink';
+import JobState from '../icons/JobState';
+import PageWithContainer from '../shared/PageWithContainer';
+import Panel from '../shared/Panel';
+import ShowMoreFooter from '../shared/ShowMoreFooter';
+
+const PAGE_SIZE = 25;
+
+type Props = {
+  agent: {
+    uuid: string,
+    name: string,
+    jobs: {
+      count: number,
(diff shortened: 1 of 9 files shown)


ticky:
Importantly, this doesn’t refresh the job list periodically like the Agent page does. Is this a problem?

ticky:
This has been updated significantly, and I would call it pretty-well ready to ship!

toolmantim:
This is awesome! Looking forward to giving it a test.

The only part that stood out for improving was the "(and 34 more)" and "Recent Jobs". Because "Recent Jobs" is a link, I thought maybe just "Jobs"?

I don't think the refreshing/staleness is a problem, as long as the jobs themselves live update?

ticky:
> as long as the jobs themselves live update?

They don’t currently, and I’m not sure what events it should be listening for. 😅

toolmantim:
Hrmm… good question! We do this on the pipelines page already, so we might need to go checking there.

Also I just realised, if we don't update the data live, can people at least refresh the page to see the jobs status change, or does that break pagination etc?

ticky:
Pagination is not kept live between reloads, no. Given that this only shows finished builds, it shouldn’t make much difference as any build on a page other than the first would hypothetically be quite old

ticky:
Yeah, sorry, I was right the first time. It’s only finished builds.

ticky:
We also don’t currently publish any agent, or job-specific push events. That’s… probably for the best.

The `build:created` and `build:updated` events only include the IDs for the objects the affect directly. this means in order for the job history page to determine that it should fetch new jobs, it would need to know that the build has or had jobs which have or are running on this agent. Currently, this could only be determined by fetching the build’s job list in its entirety from the server, which seems less than ideal.

Instead, I could take the “refresh periodically” approach the main agent page uses, which seems “good enough for now”.

toolmantim:
Can we fetch the build UUID for each job in the GraphQL query? That way we know what jobs we're showing (and their build ids), and we can listen to build:updated events for the build ids we're showing?

ticky:
We can, but that only means jobs added to builds whose jobs we’ve already been looking at will trigger a refresh. New or unblocked builds wouldn’t count.

toolmantim:
Ah yeah! Hrmmm. Doesn't `agent:updated` get fired though when a new job is assigned to the agent, which would handle those cases?

ticky:
Nope, we don’t have any agent events as far as I can see!

ticky:
Okay this has been updated to use a three-second timer. It also ships a minor update to the GraphQL endpoint which lists the currently running job in the list of jobs, rather than restricting it to those not currently being run!

PR #388, opened 2017-11-20
Add decoration to test failure annotations

jules2689:
What this does
---
Makes the annotation titles more obviously clickable:

before (with the text selection cursor):
image

after (with the hand pointer cursor on hover): 
image

cc @toolmantim @keithpitt

### DIFF
--- a/app/css/annotation-body.css
+++ b/app/css/annotation-body.css
@@ -8,6 +8,8 @@
 
 .annotation-body details summary {
   outline: none !important;
+  cursor: pointer;
+  text-decoration: underline;
 }
 
 .annotation-body details[open] summary {


ticky:
That’s weird, these “details” elements are supposed to have some kind of little disclosure triangle on them. May I ask what browser/os combo you were seeing this in?

jules2689:
Firefox Quantum Version 57.0 (64 bit)

ticky:
Aha! I’ll have a dig about with the new shiny ‘fox, then to double check what’s going wrong :) Thanks!

ticky:
Looks like this is a bug that existed in earlier versions, too. Firefox seems to implement a newer version of the spec for that element, which means our ancient copy of normalize.css is resetting it to a value which renders incorrectly in Firefox exclusively. Ouch!

jules2689:
Interesting! Well, at least the issue is known now :)

ticky:
Looks like this is now fixed. Sorry for the trouble, and thanks for the input :)

jules2689:
No problems! Thanks for the quick fix!

PR #391, opened 2017-11-23
Load bass early so our rules override

sj26:
I was trying to use the `caps` class and add `letter-spacing-0` to reset the letter spacing but it didn't work because the bass rules are later and the cascade made me cry.

So maybe these belong up here, just after the reset?

### DIFF
--- a/app/css/main.css
+++ b/app/css/main.css
@@ -1,4 +1,7 @@
 @import 'reset';
+
+@import 'basscss';
+
 @import 'btn';
 @import 'btn-primary';
 @import 'btn-outline';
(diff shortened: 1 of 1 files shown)


ticky:
Hmmm, this actually breaks a bunch of existing layout. I suspect there’ll be a better way to fix this. Do you have a branch with what you were trying to achieve on it?

master right now:

screen shot 2017-11-24 at 12 52 29 pm

this PR:

screen shot 2017-11-24 at 12 52 46 pm

sj26:
I'll close this — I don't remember what I was trying to do.

PR #395, opened 2017-12-04
Use moment duration format for duration ... formatting

sj26:
One of our customers noticed their job which had taken 2 mins 11 secs (131 seconds total) was being displayed as running for "1 minute, 11 seconds". It seems our duration formatter is a bit tricky and wasn't quite accurate all of the time.

This replaces our home-grown duration formatter with [moment-duration-format](https://github.com/jsmreese/moment-duration-format). Sorry @ticky! It seems to handle most of the edge cases we're trying to catch in our formatting, and seems more correct in every case I've examined. Some of the durations we had in test snapshots didn't seem quite correct, and this makes them more correct.

A few differences: this formatter seems more casual with rounding, and may round down in some cases. But it has seemed reasonable in every case I've looked at.

The only strangeness is: `moment.duration(1.75).format("s[secs]") // => "2 sec"`. It seems to be a rounding bug where the value is rounded but the unit name consideres a floored value. I've filed an issue upstream: https://github.com/jsmreese/moment-duration-format/issues/78


ticky:
> This replaces our home-grown duration formatter with moment-duration-format. Sorry @ticky!

No, this is good! I hadn’t found a thing which did what we needed, and I’m very glad such a thing now exists :)

app/components/shared/__snapshots__/Duration.spec.js.snap
```diff
@@ -60,15 +52,15 @@ exports[`Duration Duration.Full renders as expected 7`] = `
...
+  1 week, 6 days, 16 hours
 </span>
 `;
 
 exports[`Duration Duration.Full renders as expected 8`] = `
 <span
   className="tabular-numerals"
 >
-  8 weeks, 2 days, 20 hours
+  9 weeks, 0 days, 19 hours
```
ticky: Hmm, that’s annoying that it doesn’t remove zeroed values :(

app/lib/date.js
```diff
@@ -56,126 +57,24 @@ export function getDateString(date, withSeconds = false, withYear = true) {
...
-          .duration(lastTime.amount, lastTime.label)
-          .add(nextTime.amount, nextTime.label)
-          .as(lastTime.label)
-      );
-    }
-  } else {
-    times[0].amount = Math.round(times[0].amount);
+  if (to === undefined || from === undefined) {
+    return '0s';
   }
```
ticky: Hmm, what’s this do if it’s not guarded? This path doesn’t honour our formatter.

sj26: Oh true, I should have formatted a default param.

By default `moment(undefined)` seems to eql "current time" which means `moment(undefined).diff("some time past")` renders a negative duration.

I'll update to make it render a zero duration instead of a literal string value.

app/components/shared/__snapshots__/Duration.spec.js.snap
sj26: it seems to trim left and right but not middle.

sj26: I've got a feeling it's being worked toward a 2.0.0 release, you could probably get a feature request or contribution in before then. I'm not quite savvy enough. 😅

app/lib/date.js
ticky: Hmm, when does it get passed undefined?

sj26:
🎉

PR #398, opened 2018-01-04
Small fixes to the Audit Log list

keithpitt:
- [x] Support the new `AuditAPIContext` in GraphQL
- [x] Bump GraphQL schema
- [x] Always load new data when switching to the audit log screen


app/components/audit_log/Index.js
```diff
@@ -38,11 +38,13 @@ class AuditLogList extends React.PureComponent<Props, State> {
   };
 
   state = {
-    loading: false
+    loading: true
```
ticky: If we’re setting it on mount, do we need it to be true here?

keithpitt: This is a good question, I don't think so actually! Will fix. I think I added it here, aded it below, then forgot to remove it :P

PR #399, opened 2018-01-09
Yarn upgrade Relay, GraphQL and friends

keithpitt:
Just your regular yarn upgrade!

### DIFF
--- a/app/components/audit_log/Index.js
+++ b/app/components/audit_log/Index.js
@@ -85,14 +85,14 @@ class AuditLogList extends React.PureComponent<Props, State> {
   }
 
   renderFooter() {
-    if(this.props.relay.variables.isMounted) {
+    if (this.props.relay.variables.isMounted) {
       return (
         <ShowMoreFooter
           connection={this.props.organization.auditEvents}
           loading={this.state.loadingMore}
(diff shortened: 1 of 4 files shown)


app/components/audit_log/Index.js
```diff
@@ -87,17 +83,28 @@ class AuditLogList extends React.PureComponent<Props, State> {
       </Panel.Section>
     );
   }
+  renderFooter() {
+    if(this.props.organization.auditEvents) {
```
ticky: Is this condition not already managed by the `ShowMoreFooter`?

keithpitt: It was to get rid of this warning in console:

`Expected prop `connection` to be supplied to `ShowMoreFooter`, but got `undefined`. Pass an explicit `null` if this is intentional`

Since `this.props.organization.auditEvents` returns `undefined` if it's not been loaded yet by Relay...

ticky: Oh! Could we key it off the `isMounted` variable so it’s more explicit, then?

keithpitt: Cool, done.

PR #400, opened 2018-01-09
Yarn upgrade React, Eslint & friends

keithpitt:
This change also brings with it React 16.2, which is awesome, cause it's got awesome stuff in it!

Exhibit A: https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html


keithpitt:
Fun fact of the day:

> Special characters like the dot(.) and asterisk (*) are not special inside a character set, so they don't need to be escaped

keithpitt:
@ticky this is ready for a review.  There are still these `console.error` coming through in tests:

```
    console.error node_modules/fbjs/lib/warning.js:33
      Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.

      Check the render method of `Collapsible(function (props) {return _react2.default.createElement('pre', null, JSON.stringify(props));})`.
          in Unknown (created by Collapsible(function (props) {return _react2.default.createElement('pre', null, JSON.stringify(props));}))
          in div (created by CollapsableFormField)
          in div (created by CollapsableFormField)
          in CollapsableFormField (created by Collapsible(function (props) {return _react2.default.createElement('pre', null, JSON.stringify(props));}))
          in Collapsible(function (props) {return _react2.default.createElement('pre', null, JSON.stringify(props));})
```

...and...

```
    console.error node_modules/fbjs/lib/warning.js:33
      Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.

      Check the render method of `Labelled(function (props) {return _react2.default.createElement('pre', null, JSON.stringify(props));})`.
          in Unknown (created by Labelled(function (props) {return _react2.default.createElement('pre', null, JSON.stringify(props));}))
          in label (created by FormInputLabel)
          in FormInputLabel (created by Labelled(function (props) {return _react2.default.createElement('pre', null, JSON.stringify(props));}))
          in div (created by Labelled(function (props) {return _react2.default.createElement('pre', null, JSON.stringify(props));}))
          in Labelled(function (props) {return _react2.default.createElement('pre', null, JSON.stringify(props));})
```

But it's not blowing anything up, and everything seems to be working fine.

ticky:
>Fun fact of the day:
>>Special characters like the dot(.) and asterisk (*) are not special inside a character set, so they don't need to be escaped

I knew about this, but I’m surprised it wasn’t errorring before?

> There are still these console.error coming through in tests

If we’re not using it, should we just remove those specs entirely?

keithpitt:
@ticky thanks!

toolmantim:
Looks like this might have broken the teams page for peeps. Will try reverting for now.

PR #401, opened 2018-01-09
Yarn upgrade development dependencies (webpack, postcss)

keithpitt:
Went pretty smoothly, everything just worked! (well...we'll see after the build I suppose)


keithpitt:
@ticky I got to do some flow stuff! It's noice...

app/app.js
```diff
@@ -147,5 +147,11 @@ if (window._slanger) {
...
 // Only do the react-router gear on pages we've designated
 window["initializeReactRouter"] = () => {
-  ReactDOM.render(Routes, document.getElementById('root'));
+  const rootElement = document && document.getElementById("root");
+
+  if (rootElement) {
+    ReactDOM.render(Routes, rootElement);
+  } else {
+    throw "No #root element to render the application to!";
+  }
```
ticky: Ah, the things Flow makes us do.

I’d maybe make an `if !rootElement` shortcut and throw, so we just have an if and no else :)

ticky: Oh, also, let’s throw an Error object rather than a string (you’ll take your eye out with that!)

app/components/agent/Jobs/index.js
```diff
@@ -47,7 +47,7 @@ class AgentJobs extends React.PureComponent<Props, State> {
     loading: false
   };
 
-  _agentRefreshTimeout: number;
+  _agentRefreshTimeout: TimeoutID;
```
ticky: Ooh! Is this new? I like it. Giving names to previously opaque concepts!!!!! Awesome!

ticky:
Yeah, Flow is a nice addition, I think. I feel like it’s definitely helped me write more robust code - now if only more of our code were checked by it!

app/components/agent/Jobs/index.js
keithpitt: Yeah, it had a cry at me because we were using the wrong return type for timeouts. Much nicer!

keithpitt:
> Yeah, Flow is a nice addition, I think. I feel like it’s definitely helped me write more robust code - now if only more of our code were checked by it!

I'm ready to go all in!

ticky:
> I'm ready to go all in!

Me too, but we’ve got a lot to do to get to \~that~ point :)

PR #404, opened 2018-01-10
(Retry) Yarn upgrade React, Eslint & friends

keithpitt:
Will do some debugging in this branch to see what went wrong when it hit production.


keithpitt:
Ok, found the problem. We were doing naughty things with react, and latest react says "no no"

ticky:
That seems like a really weird thing for it to freak out about, but sure ok

keithpitt:
@ticky oh no, that's not it (my last commit). I'm still working on the *actual* fix!

keithpitt:
Spoiler: it's to do with using `ref` and accessing Dom elements from other react components and that we probably shouldn't be doing...

ticky:
Ah yes, we’re probably not doing best practices there.

keithpitt:
@ticky got it! Check out my last commit for the fix.

ticky:
I’d love to know what changed in 16.x to break this

keithpitt:
My guess it's the new rendering algorithm where the order of operations is slightly different.

keithpitt:
Maybe:

> Previously, changing the ref to a component would always detach the ref before that component’s render is called. Now, we change the ref later, when applying the changes to the DOM.

PR #405, opened 2018-01-10
Moved more packages into `vendor`

keithpitt:
I just ran an analyse of our web pack bundles and noticed the vendor one had gotten slightly out of date. This brings it back inline and there are no more `node_modules` in our app package.


webpack/config.js
```diff
@@ -97,30 +97,50 @@ var plugins = [
 var vendor_modules = [
   "autosize",
   "babel-polyfill",
+  "buffer-crc32",
   "classnames",
   "codemirror",
```
toolmantim: It'd be amazing if the codemirror stuff was a dynamic import!

keithpitt: I actually looked into it, it's like only 60kb gzipped, and our total vendor size is 390kb gzipped, not too shabby!

Relay, react and something called `core-js` are the biggest players.

keithpitt: Also React 16 is much smaller and has dropped the vendor bundle by about 20kb

PR #407, opened 2018-01-11
Yarn upgrade everything else!

keithpitt:
Not much going on here really. Some date specs broke, but I fixed.


app/lib/__snapshots__/date.spec.js.snap
```diff
@@ -18,7 +18,7 @@ exports[`getDurationString correctly handles \`full\` dates when given \`2016-05
 
 exports[`getDurationString correctly handles \`full\` dates when given \`2017-01-17T08:57:30.994-08:00\` and \`2017-02-01T17:28:11.712-08:00\` and no format override 1`] = `"2 weeks, 1 day, 8 hours"`;
 
-exports[`getDurationString correctly handles \`full\` dates when given \`2017-01-17T17:28:11.000-08:00\` and \`2017-01-17T17:28:12.712-08:00\` and no format override 1`] = `"2 second"`;
```
keithpitt: @ticky I changed this from `2 second` to `2 seconds`. The snapshot must have been wrong?

ticky: …weird, I know Sam redid a bunch of this recently

keithpitt: @sj26 do you have any memory of this place?

sj26: Yeah, this was from when we fixed duration formatting in https://github.com/buildkite/frontend/pull/395 using [moment-duration-formatter](https://github.com/jsmreese/moment-duration-format), but it had a [small bug](https://github.com/jsmreese/moment-duration-format/issues/78) when rendering fractional durations (the value 1.6 renders as "2 second" because "1.6" rounds to "2" but floors to "1" and the "second/seconds" is based on floor). Looks like they've fixed it upstream, so all good! :+1:

PR #410, opened 2018-01-18
FormRadioGroups always need onChange now

toolmantim:


keithpitt:
It does!?

toolmantim:
According to https://github.com/buildkite/frontend/commit/995d0adae43fd583d5b642b4babf343ef335ca14 and the React errors in the browser console… yeah :)

keithpitt:
Oh right! lolol.

toolmantim:
I just went and checked, and everything else specifies an onChange 👍🏼

PR #414, opened 2018-02-16
Confirm rebuilding

ticky:


keithpitt:
After much googling and asking peeps, I reckon that we should stick to no confirm for now (unless I'm totally wrong here!)

My understanding of best practice is: "assume they meant to do it"

1. If they can easily undo the action, then no confirm.
2. If it's a one-way thing and potentially dangerous (i.e. deleting data), then a confirm is a good idea.

Since they can just cancel the accidental build, then I rekon it's probably better to stick to no confirm for now.

sj26:
By that argument, job retry should also not confirm once we've got immutable jobs and cancellable jobs fully launched?

keithpitt:
@sj26 yup, I dropped confirmations from the immutable retry gear already!

PR #419, opened 2018-03-28
Backport Dropdown changes from Site

ticky:
This brings the Styled Components port of the Dropdown components used in Site into Frontend.


app/components/layout/Navigation/MyBuilds/index.js
```diff
@@ -82,9 +82,15 @@ class MyBuilds extends React.Component {
         <DropdownButton className={classNames("flex-none py0", { "lime": this.state.isDropdownVisible })} onMouseEnter={this.handleButtonMouseEnter}>
           {'My Builds '}
           <div className="xs-hide">
-            <CSSTransitionGroup transitionName="transition-appear-pop" transitionEnterTimeout={200} transitionLeaveTimeout={200}>
+            <CSSTransition
```
keithpitt: I wonder if there's a `eslint` rule we can have that forces this as the style. I personally prefer the former, but I'm happy to standardise on the latter.

ticky:
> I'd usually aim for a separate PR for the flow type upgrades

Yeah, absolutely, it’s just a thing that only bit me once I fixed things.

app/components/layout/Navigation/MyBuilds/index.js
```diff
@@ -105,14 +111,16 @@ class MyBuilds extends React.Component {
...
-    // Only render the badge if we've actually got a number to show
-    if (buildsCount) {
-      return (
-        <Badge className={classNames("hover-lime-child", { "bg-lime": this.state.isDropdownVisible })}>
-          {buildsCount}
-        </Badge>
-      );
+    // Render nothing (an empty span) if we've not got a number to show
+    if (!buildsCount) {
+      return <span />;
```
keithpitt: I think we can return `null`? (Or do we need to return a `<span>` for animations to work?)

app/components/layout/Navigation/MyBuilds/index.js
ticky: Last time I checked, none existed that didn’t mandate that you always do one or the other. I think inline makes sense for one or two short props, but as soon as it gets longer than like, 3x the component name I much prefer multiline. Makes diffs cleaner IMO.

app/components/layout/Navigation/MyBuilds/index.js
ticky: No, we need to return something that actually renders because otherwise the `CSSTransition` complains that it has no children. I honestly consider it a bug, but I think it’s a side effect of the way we’re using it.

PR #423, opened 2018-05-03
Expose the error states possible for a PipelineSchedule in the UI

ticky:
Previously, PipelineSchedules would be disabled and there was no indication other than no builds being generated, and the "next build" in the UI becoming further and further in the past that it was disabled. This exposes both the "disabled" state and failure messages in the UI, and allows re-enabling it from the error message itself.

screen shot 2018-05-03 at 6 57 21 pm

screen shot 2018-05-03 at 6 57 27 pm

This PR does not address taking ownership of a Schedule in order to re-authorise the schedule itself, but this would be a good next step.

### DIFF
--- a/app/components/pipeline/schedules/Edit.js
+++ b/app/components/pipeline/schedules/Edit.js
@@ -20,6 +20,7 @@ class Edit extends React.Component {
       commit: PropTypes.string,
       branch: PropTypes.string,
       message: PropTypes.string,
+      enabled: PropTypes.bool.isRequired,
       env: PropTypes.arrayOf(PropTypes.string),
       pipeline: PropTypes.shape({
         slug: PropTypes.string.isRequired,
@@ -55,6 +56,7 @@ class Edit extends React.Component {
                 commit={this.props.pipelineSchedule.commit}
                 branch={this.props.pipelineSchedule.branch}
                 message={this.props.pipelineSchedule.message}
+                enabled={this.props.pipelineSchedule.enabled}
                 env={this.props.pipelineSchedule.env.join("\n")}
                 ref={(form) => this.form = form}
               />
@@ -116,6 +118,7 @@ export default Relay.createContainer(Edit, {
         commit
         branch
         message
+        enabled
         env
         pipeline {
           ${Form.getFragment('pipeline')}
--- a/app/components/pipeline/schedules/Form.js
+++ b/app/components/pipeline/schedules/Form.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Relay from 'react-relay/classic';
 
+import FormCheckbox from '../../shared/FormCheckbox';
 import FormTextField from '../../shared/FormTextField';
 import FormTextarea from '../../shared/FormTextarea';
 import ValidationErrors from '../../../lib/ValidationErrors';
@@ -13,11 +14,16 @@ class Form extends React.Component {
     commit: PropTypes.string,
     branch: PropTypes.string,
     message: PropTypes.string,
+    enabled: PropTypes.bool,
     env: PropTypes.string,
     errors: PropTypes.array,
     pipeline: PropTypes.shape({
       defaultBranch: PropTypes.string.isRequired
     }).isRequired
+  };
+
+  static defaultProps = {
+    enabled: true
   };
 
   componentDidMount() {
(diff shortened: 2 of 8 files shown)


toolmantim:
This is great! It’s going to help a lot of people, I’ve had a number of confused support requests about this.

Small things that shouldn’t necessarily block this: 

* Do we have a “badge” like (small, uppercase) component for the “Disabled” label on the list view? Perhaps we put that to the left or right side of the name?
* Maybe we out a notice at the top of the list view too? That one or more schedules are currently disabled because the user is no longer a member of the org.
* Could even even just have a “re-enable” button that re-saves the schedule for them?

ticky:
This has been expanded to add an “enabled” checkbox to the schedule page, and to add a “re-enable” button for schedules which have failed (it doesn’t show for manually disabled schedules). The text in the alert has been updated to reflect this.

screen shot 2018-05-03 at 6 08 51 pm

ticky:
Clicking that button doesn’t take ownership, no. That’s also something I’d like to address, but I know that temporarily turning off the schedule is a feature we’ve been asked for, and this both addresses that and makes it clear when a schedule _is_ disabled, in either case.

ticky:
Just added a note to the list page about the errors.
screen shot 2018-05-03 at 6 28 49 pm
screen shot 2018-05-03 at 6 29 04 pm

toolmantim:
Cool okay, sounds like a good improvement either way.

How about we remove the second sentence on the list page, and just state there’s some disabled schedules? Avoids introducing new terms like “review” etc.

On the schedule page error message, if it was disabled due to an access/permissions error, and the membership hasn’t been fixed, I wonder if we should include the re-enable button or change the wording? At the moment “If their access has been restored” puts the burden on them to go and figure out whether it can be re-enabled. Or is it just a “Press re-enable and hope for the best”? If it is, then maybe we just remove “If their access was restored” sentence and let them hit the button, and get an error if it wasn’t?

ticky:
- > How about we remove the second sentence on the list page, and just state there’s some disabled schedules? Avoids introducing new terms like “review” etc.

  👌🏼

- > is it just a “Press re-enable and hope for the best”

  From my understanding, the “Re-Enable” operation will _always_ succeed. If the user who owns the Schedule still lacks build creation access, the schedule will simply be disabled again the next time it tries to run. I think this means that a “take ownership” option is going to be important.

toolmantim:
Cool, we’ll maybe then the current message is cool, but we just remove the newline?

And look into ownership stuff as a new thing?

ticky:
Taking ownership looks easy enough to do, I’d just like to look into whether it’s a reasonable step to take, and what we need to do to determine if a user can safely do so!

ticky:
OK I pinched a Bass stanza from the new team “recommended” thing, but turned it red. I’m not sure it looks good in the big size. Feel free to tweak while I’m offline, @toolmantim 😃
screen shot 2018-05-03 at 6 57 21 pm
screen shot 2018-05-03 at 6 57 27 pm

toolmantim:
No need to badge the big heading I reckon! The small one looks great 👌🏼

ticky:
Well, something needs to be present; on manually disabled schedules the badge is currently the only indication it’s disabled!
screen shot 2018-05-03 at 7 02 37 pm

toolmantim:
Oh I see! Sorry.

Can we not inherit the font size from the heading then? As in, make it the same as on the list page?

ticky:
We can, but I’ll have to do it manually I believe (it’s using bass’ `small` class which just drops it to something like .7em)!

ticky:
OK! Both schedule badges are now rendered with 12px fonts. This only changes the one on the individual schedule page.
screen shot 2018-05-04 at 2 29 09 pm

toolmantim:
Hey, looking great! A few small word things: 

* `s/no longer has access/no longer had access/`
* `s/on this Pipeline/on this pipeline/`
* Given our above discussion, let’s just remove the second sentence perhaps?

Aside from these… 🚀🎉

ticky:
Both of those messages are part of the back-end infrastructure which handles this error, and such errors already exist in this current form (they’re set when the error occurs). I can change it going forward, if that makes sense?

toolmantim:
Ahh okay. Hrmm, maybe let’s change it for errors going forward then?

ticky:
Done in https://github.com/buildkite/buildkite/pull/2709/commits/4572ab643000fbd802ccfc3d24a13d29da78afe3 !

ticky:
@toolmantim if you’re around, wanna give this a once-over? I’d love to deploy it today and get started on the “take ownership” bit

toolmantim:
Ship it! 🚀🚀🚀

toolmantim:
(Sorry I didn’t have a chance to give it more of a once over, but I reckon it looks great from the PR here!)

PR #431, opened 2018-05-30
Add System Status link

uriahcarpenter:
A quicker alternative to finding the link on https://buildkite.com/home


ticky:
Wow, good call! I totally missed that we didn’t have this in the app.

Thanks for the contribution. 😄

toolmantim:
Oh wow, how did we not already have this. Thanks @uriahcarpenter!

PR #434, opened 2018-06-01
Improve UI resilience

ticky:
This fixes a few minor oversights in our UI, where it’s not checking for the presence of objects before rendering based on them. There are a few situations where the viewer or organization information might be missing, so this updates components which would previously fail to render in this state to handle them gracefully. This state is usually encountered by staff during support, though.

### DIFF
--- a/app/components/layout/Navigation/MyBuilds/index.js
+++ b/app/components/layout/Navigation/MyBuilds/index.js
@@ -1,5 +1,6 @@
+// @flow
+
 import React from 'react';
-import PropTypes from 'prop-types';
 import Relay from 'react-relay/classic';
 import { CSSTransition } from 'react-transition-group';
 import classNames from 'classnames';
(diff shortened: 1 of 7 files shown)


app/components/layout/Navigation/MyBuilds/index.js
```diff
@@ -16,29 +17,46 @@ import CachedStateWrapper from '../../../../lib/CachedStateWrapper';
...
+  relay: Object
+};
 
+type State = {
+  isDropdownVisible: boolean,
+  scheduledBuildsCount: number,
+  runningBuildsCount: number
+};
+
+class MyBuilds extends React.Component<Props, State> {
```
sj26: This typing pattern is neat! I hadn't seen it before. 👌

ticky: It’s great! Flow is *really* flexible, but provides really valuable checks I’d love to get more coverage out of across the codebase!

sj26: 💯

PR #436, opened 2018-06-04
AWS Stack only needs agent token, now

sj26:
A while back we removed the need for an api access token from the elastic stack, but forgot to clean up the instructions in here. This simplifies the instructions and makes them look more like the others again.

image


sj26:
🙇

toolmantim:
🎉

PR #516, opened 2018-07-11
Update nvmrc to node 10

sj26:
Make sure we're using node v10 with nvm and nodenv, too.


.node-version
```diff
@@ -0,0 +1 @@
+.nvmrc
```
ticky: …is this just what a symlink looks like in git?

sj26: Yeah 😅

PR #519, opened 2018-07-17
MFA Config GUI

ticky:


app/components/user/TwoFactor/configure.js
```diff
@@ -0,0 +1,325 @@
...
+              <Panel className="orange border-orange">
+                <Panel.Section>
+                  <strong>Youʼre about to reconfigure two-factor authentication.</strong> This will replace your existing two-factor authentication applications and recovery codes.
+                </Panel.Section>
+              </Panel>
+            )}
+
+            <p>To {this.props.viewer.totp ? 'reconfigure' : 'activate'} two-factor authentication, scan this QR Code with your authenticator application.</p>
+
+            <QRCode
```
keithpitt: How much does the `QRCode` gear bloat the vendor package size? Do we need to delay load it in (like we do code mirror?) or is it kinda meh?

app/components/user/TwoFactor/configure.js
```diff
@@ -0,0 +1,325 @@
...
+      `,
+      variables: { input: {} },
+      onCompleted: this.handleCreateMutationComplete,
+      onError: this.handleCreateMutationError
+    });
+  }
+
+  componentWillUnmount() {
+    // When unmounting, if we haven't finished activating, try to delete the TOTP record
+    if (this.state.totpId && this.state.generatedTOTP && !this.state.activatedTOTP) {
```
keithpitt: Will this block an unmount? What happens if this query likes to take it's time?

app/components/user/TwoFactor/configure.js
```diff
@@ -0,0 +1,325 @@
...
+
+  handleCreateMutationError = (error) => {
+    if (error) {
+      if (error.source && error.source.type === GraphQLErrors.ESCALATION_ERROR) {
+        location.reload();
+      } else {
+        // if (error.source && error.source.type === GraphQLErrors.RECORD_VALIDATION_ERROR) {
+        //   this.setState({ errors: error.source.errors });
+        // } else {
+        alert(JSON.stringify(error));
```
keithpitt: We have the flash gear you can throw this error at (but I suspect you'd do that anyway)

app/components/user/TwoFactor/configure.js
ticky: I don’t think it’s too significant, but I’ll check!

app/components/user/TwoFactor/configure.js
ticky: Good question! This won’t block an unmount or navigation; it’s entirely async. It doesn’t really matter much if this succeeds or not (a non-activated TOTP configuration simply has no effect), this is more or less a best effort to clean up after ourselves.

This can be enhanced further (perhaps even obviated) by having a cleanup worker of some sort to delete them automatically (which we’ve discussed before, but I don’t consider high priority to implement until the rest of the shape of this is better understood!).

app/components/user/TwoFactor/configure.js
ticky: I’m going to go review what errors are actually possible in this state, and enhance the form with being able to tell you if the code is wrong.

app/components/user/TwoFactor/configure.js
ticky: This adds 9.7kb to our gzipped (actual transferred) size.
(comparing https://buildkite.com/buildkite/frontend/builds/2821 with https://buildkite.com/buildkite/frontend/builds/2818 using their bundle analyser output)

This is 1/5 the size of codemirror, and also codemirror seems to be in the main bundle anyway? Hmm.

PR #527, opened 2018-07-25
SSO2: Badges

keithpitt:
This change adds a few badges about the place.

Invitation rows now have an "Administrator" badge (it used to just be gray words), now it's a badge:

image

I've applied the same treatment here:

image

The org switcher in the nav will now check to see if can view pipelines inside it, and if it can't (because of SSO) it'll show:

image


keithpitt:
I forgot to note that if a user or an invitation has an `sso_mode` of "optional", it'll show this:

image

keithpitt:
I think this is OK!

PR #540, opened 2018-08-22
Use a caching version of moment-duration-format

ticky:
This is intended to help improve performance on pages showing large quantities of timers.

Upstream PR is at jsmreese/moment-duration-format#122

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "metrick": "0.2.0",
     "mojibaka": "1.0.2",
     "moment": "2.22.2",
-    "moment-duration-format": "2.2.2",
+    "moment-duration-format": "https://github.com/ticky/moment-duration-format.git#951b8be1d1b0424824363aa912cbf066546bf384",
     "object-assign": "4.1.1",
     "prop-types": "15.6.2",
     "pusher-js": "4.2.2",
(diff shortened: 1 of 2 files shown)


toolmantim (approved):
Awesome! 🙌🏼

I wonder how we can get a before+after measure of this optimisation? 🤔

ticky:
I don’t think we realistically can without adding some invasive data collection to our pages. If you look at the upstream PR it gives some ballpark performance numbers!

eleanorakh:
Looks good to me! 👏 I was going to ask a similar question to Tim, but it's already been answered, so 👌🏻

eleanorakh (approved):

ticky:
Went looking for a way to get some profiling info out of this. Ended up making ticky/buildkite-parallel-job-tester, and running it on localhost.

There’s a lot of data to sort through inside the profile data, but basing my profiling on filtering calls to `getDurationString`, I see the following pre-patch:

image

And the following post-patch:

image

This is sort of unscientific, but in scrolling around this a bit, I see a lot more <1ms calls with the patch, and while there are some a lot fewer >2.5ms.

Further optimisations are required, but this will drop the call time for these formatters significantly.

PR #541, opened 2018-08-23
Duration formatter optimisations

ticky:
This gives us a number of performance improvements to the Duration component.

- Duration is stored in the component state as the corresponding number of seconds. If this doesn’t change, it does not cause a re-render of the duration value.
- Duration elements will only tick each second if given an indeterminate `to` value; this means that in cases where an operation shown using it is finished, it never need re-calculate its displayed value (react lifecycle hooks for prop changes will fix the case when a `to` value _does_ then change under the component).
- Duration component has been updated to not use any of the now-deprecated React lifecycle hooks.
- Duration component doesn’t use `setInterval` anymore, instead it schedules its next update after each update has been committed.
- Here’s the big one! We now use a (polyfilled!) IntersectionObserver to only update the Durations which need it when they are actually on-screen.

So now our CPU profile on high-job-volume (tested with 500) goes from this:
cpu never gets a break oh dear

To this:
phew, the cpu gets to chill for most of each second!

This is a huge improvement for users who are doing massive amounts of parallelism!


ticky:
Notably, you may see the counters show their old values briefly after they scroll on-screen, that’s unfortunately down to how quickly react can re-render them. I think it’s a small price to pay for the huge perf improvement we’re getting out of it!

app/components/shared/Duration.js
```diff
@@ -1,20 +1,43 @@
...
 import { second } from 'metrick/duration';
-import { getDurationString } from '../../lib/date';
-
-class Duration extends React.PureComponent {
+import moment from 'moment';
+import { getDuration, getDurationString } from '../../lib/date';
+
+type Props = {
+  from?: (moment$Moment | string | number | Date | number[]),
+  to?: (moment$Moment | string | number | Date | number[]),
```
sj26: lol

ticky: I’m pretty sure we technically only ever use the `string` approach, but mirroring the moment API underlying this seemed reasonable until I can audit it properly ;)

app/components/shared/Duration.js
```diff
@@ -23,67 +46,99 @@ class Duration extends React.PureComponent {
...
+          {durationString}
+        </span>
+      );
+    }
+
     return (
-      <span className={spanClassName}>
-        {value}
-      </span>
+      <VisibilityObserver
```
sj26: This is neat!

app/components/shared/Duration.js
```diff
@@ -23,67 +46,99 @@ class Duration extends React.PureComponent {
...
-  maybeSetInterval(updateFrequency) {
-    if (updateFrequency > 0) {
-      this._interval = setInterval(() => this.updateTime(), updateFrequency);
+  maybeScheduleTick() {
+    const { from, to, updateFrequency } = this.props;
+
+    // We only want to schedule ticks if our duration is indeterminate,
+    // and our update frequency isn't zero
+    if (!(from && to) && typeof updateFrequency == 'number' && updateFrequency > 0) {
+      this._timeout = setTimeout(() => this.tick(), updateFrequency);
```
sj26: Is the churn of creating new timeouts every time negligible, compared to setting up and tearing down an interval?

sj26 (approved):
This looks amazing! Well done on the profiling and improvements. :+1:

I had a general question out of curiosity, but :shipit:

app/components/shared/Duration.js
ticky: Timeout setups are super negligible. This is actually just so that we end up with less work queued overall; an interval _can_ end up queueing up work if it takes too long, while this approach means new work is only ever scheduled after the last unit of work has been done (read: React has rendered it)

app/components/shared/Duration.js
ticky: Yeah! I was super glad to see there was an implementation and polyfill for this tech! It’s solid!

app/components/shared/Duration.js
sj26: :+1:

PR #542, opened 2018-08-24
Update flow-typed type definitions

ticky:
This isn’t really going to require a review or a deploy, this is just updating type definitions (particularly stubs!) which got a bit outdated.

(p.s. I wonder if we could add flow-typed support to @renovate-bot? ;) )


sj26 (approved):
Such diff!

PR #546, opened 2018-09-11
Allow for graceful Agent stopping

ticky:
This changes the behaviour of the existing Agents page, which treats the "Stop" button as an immediate, ungraceful, force stop. This means we don't have any user-facing way to gracefully stop an agent!

Following this change, "Stop" now gracefully stops.

If the Agent takes more than 5 (±3) seconds to stop, we then show an additional "Force Stop Agent" section below, with a warning that it'll cancel the running job (as the current warning does) and will function the same as our current behaviour.

### DIFF
--- a/app/components/agent/Show.js
+++ b/app/components/agent/Show.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Relay from 'react-relay/classic';
 import { Link } from 'react-router';
+import moment from 'moment';
 import styled from 'styled-components';
 import DocumentTitle from 'react-document-title';
 import { seconds } from 'metrick/duration';
@@ -44,6 +45,8 @@ class AgentShow extends React.Component {
       uuid: PropTypes.string,
       name: PropTypes.string,
       connectionState: PropTypes.string,
+      disconnectedAt: PropTypes.string,
+      stoppedAt: PropTypes.string,
       job: PropTypes.object,
       permissions: PropTypes.shape({
         agentStop: PropTypes.shape({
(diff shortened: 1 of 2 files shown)


lox (approved):
Nice one! 💯

app/components/agent/Show.js
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Relay from 'react-relay/classic';
 import { Link } from 'react-router';
+import moment from 'moment';
```
toolmantim: Should we just import just https://date-fns.org/v1.29.0/docs/differenceInSeconds instead of using moment? Seems to be the more minimal equivalent I seem people using, that also tree shakes nicely w/ webpack.

toolmantim (commented):
This is awesome! 👏🏼

toolmantim:
There's a few improvements I can spot, that'd be great to get in here:

* Maybe keep the "Stopping…" disabled button and section visible until the agent actually stops. So you don't have to scroll up, you can just stay there waiting for 3 sections until a new "Force Stop Agent" section appears below it?
* We can probably move the "If the agent doesn't respond..." words down to the bottom w/ the "Stopping…" button.
* In table of attributes, "Stopped" is shown in the "stopping" state… which looks inconsistent/lies because the agent hasn't stopped. Maybe we rename to "Stopped By" and rejig the sentence to "Jessica Stokes, today at 6:03 PM"

app/components/agent/Show.js
ticky: I mean, if we didn’t already have moment loaded at all times I’d say sure :P

ticky:
Looks like this change doesn’t work after all due to backend limitations. Closing for now.

PR #578, opened 2018-09-15
Webpack 4 prep: Lazy Load CodeMirrors

ticky:
This should fix up the chunking issue we’d been having with Codemirror stuff; there was an entry point which wasn’t lazy-loaded, which means none of them got to be!

Pulled out of #511.

### DIFF
--- a/app/components/user/graphql/GraphQLExplorerConsoleEditor.js
+++ b/app/components/user/graphql/GraphQLExplorerConsoleEditor.js
@@ -1,5 +1,4 @@
 // @flow
-/* eslint-disable react/prop-types */
 
 import React from "react";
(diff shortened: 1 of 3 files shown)


ticky:
_Technically_ this leaves us with two copies of CodeMirror, this will be fixed by Webpack 4.

keithpitt:
Sorry, such an epic late reply to this! Does this mean there's 2 copies of CodeMirror in the CodeMirror bundle? Or 1 copy in the CodeMirror bundle, and another copy in the app one?

ticky:
One copy each in two lazy loaded chunks, depending on which codemirror component you end up using

keithpitt (approved):

PR #589, opened 2018-09-24
Add support for iPhone X "safe area"

ticky:
This updates calculateViewportOffsets to handle the “safe area” as margins on current Apple flagship phones.


toolmantim:
🚀

PR #593, opened 2018-09-26
Add a package script so you can `[npm,yarn] run flow`

toolmantim:
👉🏼🥫


ticky:
That should’ve already worked, at least with yarn, given it’s not running with any other arguments?

PR #595, opened 2018-09-27
Add Matthew to the support dialog

matthewd:
🤳🏻

The non-`-small` images don't seem to be used anywhere, so I didn't bother adding one.

I didn't fit within the existing dialog, so I've made it wider (I've never used React, so there was some guessing involved -- separately from whether that's the thing we want to do, did I do it correctly?). By my judgement, that'll work for now (and I added enough for another 1-2 more faces), but the next step would be to force wrapping -- which is already styled to look great, apart from needing similar numbers of people per row... and really wanting the total number to be odd.


toolmantim:
🎉

PR #599, opened 2018-09-27
Make support dialog mugshots wrap in often-even rows

ticky:
I got inspired by @matthewd’s comments in #595 😅

This adds technology to the support dialog to adjust the wrapping behaviour to construct even rows. This always errs on the side of having more, more even rows:

image

If we ended up with an extra Sam (which would be impressive!):
image

And if we get to 15, we’ll wrap to three rows, but it’ll just keep on going. Maybe we’ll have to adjust the size eventually, but this seems to give us plenty of mileage!
image

As always, this will wrap “naturally” once the screen gets too small, we don’t seek to control that.

### DIFF
--- a/app/components/layout/Navigation/support-dialog.js
+++ b/app/components/layout/Navigation/support-dialog.js
@@ -7,8 +7,13 @@ import Button from '../../shared/Button';
 import Dialog from '../../shared/Dialog';
 import Emojify from '../../shared/Emojify';
 
-const IMAGE_PADDING_HORIZONTAL = 6;
-const IMAGE_PADDING_VERTICAL = 9;
+const IMAGE_DIAMETER = 70;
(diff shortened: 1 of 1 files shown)


toolmantim (approved):
📈

plasticine:
Radness — love it

matthewd:
Sorry, I'd missed that this was targeting my branch, so I've re-pointed it to master.

IMO it'd be ideal if this could force staggering when the count is even, but that seems pretty tricky to achieve (e.g. by inserting a transparent "ghost" after my face in the first image above) without breaking wrapping in narrower windows.

ticky:
I was always aiming for this popup to handle staggered/not-staggered layouts just fine, and I think it looks okay in both permutations, but we can certainly change how we handle that. :)

toolmantim (approved):

PR #606, opened 2018-10-08
Relay Modern Spike

plasticine:
Just putting this up for sharing mostly. 

While exploring around the frontend codebase to familiarise myself a little, and on the way I thought I’d see how tackling converting a leaf-node component/fragment from Relay Classic to Modern looked like for us.

Turns out it’s not _too_ massive (granted, I started with literally the most trivial component I could find).

This patch also starts tracking Relay compiler `__generated__` files, and using the generated Flow types within, eg:

screen shot 2018-10-08 at 2 35 02 pm


keithpitt:
Nice! Yeah, the conversion is pretty straight forward, there's just a few tricky places where we delay loading in data. We have a trick in places where we only get data if there's a `hasMounted` type relay variable, meaning we can delay loading in data until the component has been rendered, instead of blocking the entire page.

If you find any components that are easy to change to modern, be my guest!

.gitignore
```diff
@@ -19,6 +19,3 @@ coverage
 
 # Webpack bundle analysis
 bundle-analysis
-
-# Babel generated files (from relay-compiler)
-__generated__
```
keithpitt: I do worry that we're going to make PR diffs spew a bunch of useless stuff for each change we make. I think I remember this being best practice (committing the stuff into the repo) - what happens if we just re-create them on first compile?

.flowconfig
```diff
@@ -1,6 +1,5 @@
 [ignore]
-.*/node_modules/.*
-.*/__generated__/.*
+.*/node_modules/config-chain/test/broken.json
```
keithpitt: What's this about? :)

.gitignore
plasticine: In practice I’ve not found it to be too much of an issue in previous projects (it’s rare for them to churn), though there will obviously be somewhat more noise, but mostly just around actual add/edit of queries/mutations — which is kinda useful?

The disadvantage of not committing them along is that we’re going to need to generate them again before compiles / flow checking can happen because we consume the generated types in the code;

https://github.com/buildkite/frontend/blob/2c8481aeb223101fa2fcc18c388a62d1b6226f97/app/components/team/Labels.js#L8-L12

.flowconfig
plasticine: Good point — will annotate! 👍

plasticine: See 13e5003

.gitignore
keithpitt: Seems legit

PR #608, opened 2018-10-08
Add Storybook

toolmantim:
This adds a `yarn storybook` command, where we can design bits of UI a bit more easily (and more easily experiment with frontend independently of backend… because it has all the CSS scaffolding set up)

image

I've started it off with a bunch of stories in a single directory, to try to get into the swing of it, but the structure/organizing of the stories is completely up for grabs.  For example:

- Should they live alongside the actual components (just with `.stories.js`)?
- Should the story nesting match the filesystem nesting? (tho I found `shared` a bit weird when I tried it)


plasticine:
This is rad — awesome. Keen to spin it up and have a poke around with the Sketch stuff...time to install Sketch!

ticky (approved):

PR #611, opened 2018-10-09
Regenerate recovery codes button

eleanorakh:
This wires up the bones of the code to handle recovery code regeneration for MFA. While the UI doesn't look great, it's likely there's going to be some more overarching UI changes soon so it's important to have the regeneration functionality in place.

Big thanks to Justin for helping me with the final bits of the mutation today 🙏✨


ticky:
I just noticed this is on the “Configure” page - those recovery codes are always new already, and also aren’t supposed to apply until you’ve activated the OTP! This ought to be on the “View recovery codes” dialog.

PR #613, opened 2018-10-09
Replace shuffle-array by n-shuffle

rap2hpoutre:
I replaced **shuffle-array** by **n-shuffle**. Motivations:
 - **n-shuffle** is only 223 bytes (**shuffle-array** is 1980 bytes)
 - **n-shuffle** is fast, has only the function required by your component, and has less code
 - ... I just created **n-shuffle** today, so it's _self-promotion_ (sorry), that's why you can close it without a comment!

Feel free to close if you find it useless!


keithpitt:
@rap2hpoutre thanks for the PR, we're pretty happy with `shuffle-array` for the moment. Take care!

PR #615, opened 2018-10-10
Relay / GraphQL

plasticine:
OK, one more time with feeling! 😛

This is a follow up to #606 to address concerns raised around storing generated Relay files inside the repo, under version control — as discussed in the engineering catch-up this morning. This is a pretty critical part of the graphql / relay workflow, and I’d really like to get it sorted out. 😄

This patch reworks the strategy introduced in #606 for the GraphQL server & Relay client workflow to remove the need to store the compiled our Relay queries and types ([ref](https://facebook.github.io/relay/docs/en/graphql-in-relay.html#relay-compiler)) in the repo by generating them on the fly instead.

This patch _does not_ remove the (pre-existing) generated GraphQL schema dump (`app/graph/schema.json`). That will have to wait for another day; changing it massively complicates our current testing infra / docker setup. 🙃 This patch _does_ automate the dumping of the schema however ([previously a manual process](https://github.com/buildkite/frontend/commits/6d7ccf57fd097f9e278f501f332e997d9934c616/app/graph/schema.json)), meaning that the schema dump in the frontend should actually be indicative of the actual server state, and will drift a lot less.

This patch introduces a new process that provides Relay compilation outside of Webpack. This is required as having a webpack handle it creates a chicken/egg problem when attempting to consume the flow types in our source code from the generated code. Compilation would never succeed if the generated files were not present, meaning that the generated could never be compiled into place. 🥚 LOL.

This new process is a small(ish) [node script](https://github.com/buildkite/frontend/pull/615/files#diff-0d20f14b4c50e5c75965bb7910e5590e) in the `frontend` repo that is booted with the rest of the development environment via the main `Procfile`.

It sets up watches on the filesystem for changes to both the JS source code, and also the Ruby GraphQL source code. It does roughly this;
- **When the JS code changes:**
  - (re)compile the generated Relay files using the JS source and the schema dump.
- **When the Ruby code changes:** 
  - sync a fresh copy of the introspection schema to disk via pulling it from the graph HTTP endpoint.
    - _PS: is this a problem storing this in a public repo? Are there any sensitive fields here?_ 🤔
  - (re)compile the generated Relay files using the JS source and the schema dump.

I’ve pinged a few of you for thoughts on this. While the script has a bit going on it’s pretty much what’s going to be required IIRC if we don’t want the generated files residing in the repo.

screen shot 2018-10-10 at 6 41 27 pm


script/watch_graph
```diff
@@ -0,0 +1,183 @@
+#!/usr/bin/env node
+/* eslint-disable */
```
ticky: 🤔

ticky (changes requested):
Seems legit! Can we have the `watch_graph` script follow our eslints, though please? 😄

plasticine:
> Seems legit! Can we have the watch_graph script follow our eslints, though please? 😄

Sure.

script/watch_graph
plasticine: ✅ done!

ticky (approved):

webpack/config.js
```diff
@@ -40,13 +39,6 @@ const chunkFilename = IS_PRODUCTION ? "[chunkhash].chunk.js" : "[name].chunk.js"
 const devTool = IS_PRODUCTION ? "source-map" : "inline-cheap-source-map";
 
 var plugins = [
-  // Forked version of the RelayCompilerWebpackPlugin that supports
-  // relay-compiler 1.1.0
-  new RelayCompilerPlugin({
-    schema: path.join(__dirname, '../app/graph/schema.json'),
-    src: path.resolve(__dirname, '../app')
-  }),
```
sj26: So I thought we avoided running relay-compiler manually by using this relay compiler webpack plugin, which webpack dev server and the build process run to create the required generated files. Is the trouble that these files aren't generated in time to be consumed by our own files?

sj26: Also, does webpack 4 address any of this?

sj26: I love javascript. 😭

sj26 (approved):

PR #625, opened 2018-10-12
FYI: Ensure that relay always compiles before a build

plasticine:
This fixes an issue that I’d missed where buildkite/buildkite will run a build, this PR ensures that relay is always compiled out before a build runs.


sj26:
Wow, does npm/yarn magically run “pre*” commands?

plasticine:
Yep, also “post”

PR #630, opened 2018-10-15
Some extra jest config for better living

plasticine:
Mainly just allows tests in `__tests__` dirs too, as is jest default.


ticky:
I’m not a huge fan of putting tests in a subdirectory because it makes them harder to jump to from ⌘P :)

PR #632, opened 2018-10-15
[WIP] MFA consumed recovery codes

plasticine:
@eleanorakh PRing this so we can track it 🙃

Oh, and this has https://github.com/buildkite/frontend/pull/633 in it for it to work!


app/components/RecoveryCodeList/index.js
```diff
@@ -2,27 +2,32 @@
...
-      >
-        {this.props.recoveryCodes.codes.map((code) => (
-          <li key={code}>
-            <code className="monospace h2">{code}</code>
-          </li>
-        ))}
+      <ul className="list-reset center" style={{ columns: 2 }}>
+        {
+          this.props.recoveryCodes.codes.map(({ code, consumed }, index) => (
+            <li key={index}>
```
plasticine: TODO: This really needs to be `id` not the index!

app/components/RecoveryCodeList/index.js
```diff
@@ -2,27 +2,32 @@
...
+              <code
+                className="monospace h2"
+                style={{ textDecoration: (consumed ? 'line-through' : 'none') }}
```
plasticine: TODO: This is pretty janky — probably should be a proper component to allow us to have the hover behaviour and to tidy up the styling.

app/components/user/TwoFactor/RecoveryCodes/RecoveryCodeDialog.js
```diff
@@ -0,0 +1,206 @@
...
+            ? this.renderTotpRecoveryCodes()
+            : this.renderGenerateTotpRecoveryCodes()
+          )}
+        </div>
+      </Dialog>
+    );
+  }
+
+  // TODO: Dry this up...
+  renderGenerateTotpRecoveryCodes() {
```
plasticine: The high level of duplication in `renderGenerateTotpRecoveryCodes` & `renderTotpRecoveryCodes` is not ideal — I think it would be worth maybe refactoring up the dynamic `generateRecoveryCodes` state stuff into a component that lives above this one and then have a simpler UI for actually rendering the dialog content.

app/components/user/TwoFactor/RecoveryCodes/RecoveryCodeDialog.js
```diff
@@ -0,0 +1,206 @@
...
+            outline={true}
+          >
+            Copy Recovery Codes
+          </Button>
+        </CopyToClipboard>
+      </React.Fragment>
+    );
+  }
+
+  generateRecoveryCodes() {
```
plasticine: Might be worth pulling this whole function & mutation up into the parent component...?

app/components/user/TwoFactor/RecoveryCodes/RecoveryCodeDialog.js
ticky: `renderGenerateTotpRecoveryCodes` no longer needs to exist, as it isn’t possible anymore for a user to have TOTP active but no recovery codes!

eleanorakh:
This is awesome @plasticine

app/components/user/TwoFactor/RecoveryCodes/RecoveryCodeDialog.js
```diff
@@ -0,0 +1,204 @@
...
+  state = {
+    generatingRecoveryCodes: false,
+    generatedRecoveryCodes: false,
+    recoveryCodes: null
+  };
+
+  componentDidMount() {
+    // If we have a TOTP, but no recovery codes for it, immediately go generate some recovery codes.
+    if (this.props.totp && !this.props.totp.recoveryCodes) {
+      this.generateRecoveryCodes();
```
eleanorakh: @ticky We were having a look at this while talking about a refactor earlier today and we're not sure if it's still necessary? What do you think?

ticky: That’s correct! This state should no longer be reachable. I was originally going to make the transition to recovery codes self-service but then Keith did some database magic instead :)

plasticine:
@eleanorakh I think we can close this out now!

PR #633, opened 2018-10-15
Some workflow tweaks

plasticine:
Commit log has the changes, but to repeat here with some more detail;

- Allow Webpack to resolve module paths that are within `app/*` to be resolved in imports using a module path, not a relative one. This makes refactoring code around on the filesystem a _lot_ easier as you don’t have to wrestle with relative paths, it also has huge gainz for find-replacing, etc etc. Of course both relative and modules imports continue to work, but using module imports for all our code is so much easier on the brain.🧠✨
  - **Before** https://github.com/buildkite/frontend/blob/9b027afcc6f41ddd14e38b436cf5814e320aae76/app/components/user/TwoFactor/RecoveryCodes/dialog.js#L7-L11
  - **After** https://github.com/buildkite/frontend/blob/c4dc7bd0794ceae995c040e3f2885cd94a76d2ea/app/components/user/TwoFactor/RecoveryCodes/RecoveryCodeDialog.js#L6-L9
- Allow flow to resolve paths in `app/` (really just the flow side of the above)
- Don’t error on unresolved or missing modules / imports. Flow already does this and usually does a better job. No point throwing two errors 🤷🏼‍♀️
- Allow unused vars if prefixed by `_` — this is a nice to have for times you want to annotate / explain what unused positional args are.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -78,7 +78,9 @@ rules:
   no-new-require: error
   no-trailing-spaces: error
   no-unsafe-negation: error
-  no-unused-vars: error
+  no-unused-vars:
+    - error
+    - varsIgnorePattern: ^_
   no-useless-rename: warn
   no-var: warn
 
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -120,6 +120,7 @@ rules:
 
   import/extensions: error
   import/unambiguous: off
+  import/no-unresolved: off  # Covered by flow
 
   react/display-name:
     - error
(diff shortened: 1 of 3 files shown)


ticky:
I’ve been meaning to add the module path alias since we landed Site, which uses this approach extensively. It’d also more closely match what the `Webpack.require` bridge uses for component names!

plasticine:
@buildkite/javascript I’m going to put this in as I reckon it’s super handy to have! If there are any dramas it’s pretty easy to revisit.

PR #635, opened 2018-10-15
The great Babel 7.0 migration

ticky:
Updates us to Babel 7.0, keeping behaviour otherwise the same.

One thing we lose in this transition is the React-specific hot reloading infrastructure, which I don’t believe worked properly as of Webpack _3_, let alone version 4. There are new ways to accomplish it which we can explore in a future PR.

The MyBuilds fixes will also be part of completion of #634.


.babelrc
```diff
@@ -1,17 +1,30 @@
...
-    "react",
-    ["env", {"modules": false}],
-    "stage-0"
+    "@babel/flow",
+    "@babel/react",
+    ["@babel/env", {"modules": false}]
   ],
   "plugins": [
-    ["relay", {"compat": true, "schema": "app/graph/schema.json"}]
+    ["relay", {"compat": true, "schema": "app/graph/schema.json"}],
```
plasticine: Won’t need this anymore!

package.json
```diff
@@ -27,6 +27,27 @@
...
+    "@babel/plugin-proposal-numeric-separator": "^7.0.0",
+    "@babel/plugin-proposal-optional-chaining": "^7.0.0",
+    "@babel/plugin-proposal-pipeline-operator": "^7.0.0",
+    "@babel/plugin-proposal-throw-expressions": "^7.0.0",
+    "@babel/plugin-syntax-dynamic-import": "^7.0.0",
+    "@babel/plugin-syntax-import-meta": "^7.0.0",
+    "@babel/polyfill": "^7.0.0",
+    "@babel/preset-env": "^7.0.0",
+    "@babel/preset-flow": "^7.0.0",
+    "@babel/preset-react": "^7.0.0",
```
plasticine: lol babel

plasticine (approved):
LGTM

package.json
ticky: I mean I consider this on us for using `stage-0` ;P

.babelrc
ticky: Oh, really? Can it just be dropped?

plasticine: Oh wait — sorry, I was thinking about webpack when I wrote this lol. Excuse me, I haven’t had a coffee yet 🙄

package.json
plasticine: 😄🙃

ticky: There are _almost certainly_ some things we can drop from this list, but this is what we had so I’m keeping them for now. We can make more informed decisions about it in future.

PR #672, opened 2018-10-26
Fix undefined data race condition from Relay @include field

plasticine:
Need to guard against `undefined` when accessing `viewer.scheduledBuilds` & `viewer.runningBuilds`. Because we’re selectively `@include`-ing these fields there appears to be a potential race condition around them being present.

screen shot 2018-10-26 at 3 43 21 pm


lox:
🚢

PR #678, opened 2018-10-29
Codemod for re-writing all our relative paths to ones resolved by “app/”

plasticine:
Had this lying around from a previous project. I won’t run it just yet as it touches ~134 files, but I’ll add it here for visibility and look to run it soon and then follow up with another PR.

You can test it out locally using the following command from the root of the Frontend repo.

```
jscodeshift app/ --transform codemods/no-relative-required.js
```

### DIFF
--- a/codemods/no-relative-required.js
+++ b/codemods/no-relative-required.js
@@ -1,0 +1,42 @@
+import path, {resolve, normalize, relative, dirname, parse} from 'path';
+
+export const parser = 'flow';
+
+function isRelativePath(path) {
+  return path.slice(0, 3) === '../';
+}
+
+// This codemod rewrites relative imports to 'absolute', module style imports.
+export default function removeRelativeImports(file, api, options) {
+  const {dir: fileDir} = parse(file.path)
+  const {jscodeshift} = api;
+  const printOptions = options.printOptions || {quote: 'single'};
+  const root = jscodeshift(file.source)
+  const nodesToUpdate = new Set();
+
+  root
+    .find(jscodeshift.ImportDeclaration)
+    .find(jscodeshift.Literal)
+    .filter(node => isRelativePath(node.value.value))
+    .forEach(path => nodesToUpdate.add(path));
+
+  root
+    .find(jscodeshift.ExportNamedDeclaration)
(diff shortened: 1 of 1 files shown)


plasticine:
> I’ve run this locally and it seems legit! This has me wondering, though, if we should drop the app prefix altogether, and make the top level under that? That would then more or less match the names used in our Webpack bridge!

@ticky Rad! 😀 Also thanks for fixing eslint! Hrmm — yeah I guess we could, though I wonder what it’d buy us—aren’t we looking to move away from that bridge code anyhoo? 🤔

I reckon for my💰 if we’re thinking about some file system changes I’d be interested in flattening from the other end a bit first. As old mate Abramov says;

> You’re overthinking your “file structure”. Create a few top-level folders and put all stuff into them. Don’t go deeper.
>
> — https://twitter.com/dan_abramov/status/1054120249870692353

I think that, as well as some clearer module naming would yield a bunch of wins; better grep-ability, components easier to styleguide, less coupling & “private” components, etc. I reckon maybe that might give us more bang for our buck—but then I’m still getting my head around the webpack/rails split—what d’you reckon?

ticky:
Oh, I didn’t mean moving the files, I meant just treating `app` as the root, as it’s really only there to split the repo-level things from the app itself! i.e. `'app/components/shared/Panel'` becomes `'components/shared/Panel'` (which it’s also called in the Webpack bridge), but doesn’t move on-disk.

The Webpack bridge is definitely something we want to move away from, but I don’t think it’s going away soon enough not to take the opportunity for import name parity here!

ticky:
I’m also here for flattening the component hierarchy as well, but this wasn’t what I was getting at! :P

plasticine:
@ticky Ahhh — sorry for my misunderstanding, I get you now! ☕️😅 Yeah I guess we could totally do that, the only drawback would be having to manage > 1 resolve aliases for webpack and flow, right?

You’d end up with a lot of entries in here for example; 

https://github.com/buildkite/frontend/blob/f2d1a984d3dfe8efc847096b1b469520971dc3fe/.flowconfig#L17

PR #681, opened 2018-10-29
Fix loading of CSS variables in JS

toolmantim:
Undoes the hardcoding in #679 by fixing the CSS variable importing, by only importing `variables.css` file, which is probably clearer and faster anyhow.

### DIFF
--- a/app/components/audit_log/row.js
+++ b/app/components/audit_log/row.js
@@ -14,7 +14,7 @@ import AuditLogDrawer from './Drawer';
 
 import { indefiniteArticleFor } from '../../lib/words';
 
-import cssVariables from '../../css';
+import cssVariables from '../../css/variables';
 
 const TransitionMaxHeight = styled.div`
   transition: max-height 400ms;
--- a/app/components/shared/TabControl/Tab.js
+++ b/app/components/shared/TabControl/Tab.js
@@ -7,6 +7,8 @@ import styled from 'styled-components';
 
 import Badge from '../Badge';
 
+import cssVariables from '../../../css/variables';
+
 import { formatNumber } from '../../../lib/number';
 
 const ACTIVE_CLASS_NAME = 'active';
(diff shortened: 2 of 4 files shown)


ticky:
Are we anywhere near being able to run Percy against frontend to catch stuff like this?

plasticine:
Ha — TIL about `css-variables-loader`. 🙄 Thanks for making the actual fix @toolmantim 👏🏼

> Are we anywhere near being able to run Percy against frontend to catch stuff like this?

Yus! I’d love to get to this point very much. Step one is progressing @toolmantim’s work on storybook so that it’s a core part of our workflow. Without it being the reference we use it’s not much use validating it!

toolmantim:
Adding Percy to the feature specs and storybook was exactly my thought too @ticky!

ticky:
I was more thinking pointing Percy at a bunch (all?) frontend-powered routes, so we can validate changes, but storybook is good too

toolmantim:
Yeah, that’d have to be via feature specs tho I think?

plasticine:
FWIW/my $0.02 — every time I’ve done screenshot diffing/validation the value decreases dramatically and work increases as soon as you try and test larger, UI with any sort of dynamic content. You just end up with noise from false positives, it becomes a broken window. 🤷🏼‍♂️

toolmantim:
I'd agree, if it made the whole build fail on diff. But w/ Percy (with the GitHub PR validation flow) it just uses GH commit statuses and super streamlined UI for just hitting "yeah cool" and changes the commit status to green.

ticky:
Oh, I was figuring it was more for review - I admit I haven’t actually seen it in action - but if we could just have it flagging that there is a diff vs marking diff as failure that’d be good

I think the thing that worries me about it being few, specific places is that there’s so much surface area which _isn’t_ covered by any feature tests

toolmantim:
I think I just realized what you might have already @plasticine, when you mentioned “dynamic content” — most of our feature spec fixtures have dynamic content in them, which would cause diff breaks 🤔

plasticine:
@toolmantim Hehe, yep — you have to have it run with fixtures otherwise you’ll be driven insane by false positives. That’s where highly componentizing you UI, and building a library where all states are possible (i.e storybook) is perfect for visual testing. Remember, components don’t have to be just a button on whatever. Having a story for a whole formset is possible to! It’s components all the way down! 🐢🐢🐢🐢

PR #692, opened 2018-10-31
Allow overriding public path dynamically

sj26:
This uses a special webpack variable, `__webpack_public_path__`, to optionally allow dynamically overriding the asset host on the fly at runtime using `window._frontendPublicPath`:

https://webpack.js.org/guides/public-path/#on-the-fly

This means we can use the same assets baked into the Docker image for testing by hosting them in a static directory and overriding the public path, as well as the asset cdn on the production assets, or download and use them locally in development too. I also want to try putting up a cloudfront version of  `buildkite.com` with a static `/_frontend` (and `/assets`, etc) so we can same-domain all the assets and utilise http/2.

It has to be in a separate module imported at the top of the entry point to be effective due to load and execute ordering.

After this PR, the next step will be to get the backend exporting `FRONTEND_HOST` as `window._frontendPublicPath`:

https://github.com/buildkite/buildkite/compare/frontend-dynamic-public-path

Then we can remove knowledge of `FRONTEND_HOST` from webpack entirely:

https://github.com/buildkite/frontend/compare/dynamic-public-path...remove-frontend-host

### DIFF
--- a/app/public-path.js
+++ b/app/public-path.js
@@ -1,0 +1,9 @@
+/* global __webpack_public_path__:true */
+/* exported __webpack_public_path__ */
+
+// Allow overriding the webpack public path
+// Must be done in a module so that it is executed before all other imports.
+// See https://webpack.js.org/guides/public-path/#on-the-fly
+if (window._frontendPublicPath) {
+  __webpack_public_path__ = window._frontendPublicPath;
+}
--- a/app/app.js
+++ b/app/app.js
@@ -1,5 +1,7 @@
 // @flow
 
+import './public-path';
+
 import React from 'react';
 import Relay from 'react-relay/classic';
 import reactRender from './lib/reactRenderer';


sj26:
I don't know how to fix eslint :-(

app/public-path.js
```diff
@@ -0,0 +1,6 @@
+// Allow overriding the webpack public path
```
ticky: I think you just need to tell it that this is an expected global!

```suggestion
/* global __webpack_public_path__ */
// Allow overriding the webpack public path
```

sj26: Sounds like eslint needs a "webpack" environment which understands the __webpack_*__ globals:

https://eslint.org/docs/user-guide/configuring#specifying-environments

sj26: This fixes the read-only thing:

```js
/* global __webpack_public_path__:true */
```

But it still complains about:

```
  1:11  error  '__webpack_public_path__' is defined but never used. Allowed unused vars must match /^_/  no-unused-vars
```

Despite the name matching `/^_/`!

sj26: Figured it out! I need an `/* exported __webpack_public_path__ */` as well!

ticky: Aha! Sorry for the bumsteer!

sj26:
Getting this in.

PR #698, opened 2018-11-01
"Generate New Recovery Code" button shift

keithpitt:
Moved it out of the activation flow and into the dialog itself.

Also switched up the rendering of the dialog so it can animate open.

nov-01-2018 15-38-36


keithpitt:
Yolo merge, I'm sure it's fine :)

PR #709, opened 2018-11-07
Rewrite relative module requirements with codemod

plasticine:
This PR runs the codemod from #678;

```
jscodeshift app/ --transform codemods/no-relative-required.js
```

### DIFF
--- a/app/components/shared/FormYAMLEditorField/codemirror.js
+++ b/app/components/shared/FormYAMLEditorField/codemirror.js
@@ -1,6 +1,6 @@
 export default from 'codemirror';
 
-import 'app/css/codemirror.css';
+import 'app/css/codemirror';
 
 import 'codemirror/addon/hint/show-hint';
 import 'codemirror/addon/comment/comment';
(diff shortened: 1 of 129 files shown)


app/components/shared/FormYAMLEditorField/codemirror.js
```diff
@@ -1,6 +1,6 @@
 export default from 'codemirror';
 
-import 'app/css/codemirror';
+import 'app/css/codemirror.css'; // eslint-disable-line import/extensions
```
ticky: Should we [allow CSS extensions](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md) explicitly, rather than using `disable-line`?

plasticine: Yeah, was pondering that a bit too — I kinda feel like I like the extra terseness when importing non `*.js` files, what d’you reckon?

plasticine: Have done. 🙃🕺🏼

PR #714, opened 2018-11-21
End-user organization icon uploads

ticky:


keithpitt (approved):

app/lib/AssetUploader.js
```diff
@@ -104,11 +104,14 @@ class AssetUploader {
...
-        this._finalizeFileUpload(upload, file);
-      } else {
-        this.options.onError(new AssetUploaderError("There was an error uploading the file. Please try again."));
+      if (!response.ok) {
+        throw null;
       }
+      this._finalizeFileUpload(upload, file);
+    }).catch(() => {
+      this.options.onError(
+        new AssetUploaderError("There was an error uploading the file. Please try again.")
```
keithpitt: I wonder if we should include the original exception in the `AssetUploaderError` thing so clients can inspect it? (doesn't matter really, just a side thought)

ticky: As like, a property or something?

keithpitt: Yeah - it may not be possible (and it was just a random idea, happy for you to ignore)

PR #719, opened 2018-11-22
Feature community forum in footer

plasticine:
screen shot 2018-11-22 at 3 18 21 pm


toolmantim:
What do you think of just “Forum” for the footer?

keithpitt:
(I'd vote for just Forum as well...)

plasticine:
@toolmantim @keithpitt Yepo, kinda anticipated that but was keeping it consistent with public site as I wasn’t super sure about the naming we’d want. +1 for “Forum” too though!

toolmantim:
✈️

PR #720, opened 2018-11-22
Remove webpack config sha prefix

sj26:
This needs to be removed to do variable frontend paths. It's part of the public path which is being replaced on the fly, and we don't store this prefix anywhere else useful.

I remember we did this because webpack config changes weren't changing asset names and caches weren't expiring correctly. I think maybe something to do with the way the md5 extension thing we were using at the time was generating hashes. But the [chunkhash] we use now is based on the content of the asset and avoid that whole problem, so I'm pretty sure we don't need it any more.

Can anybody think of a reason we need to keep it?

### DIFF
--- a/.buildkite/steps/webpack.sh
+++ b/.buildkite/steps/webpack.sh
@@ -1,14 +1,6 @@
 #!/bin/bash
 set -euo pipefail
 
-# Add the SHA1 sum of the webpack file to the host path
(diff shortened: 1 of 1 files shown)


ticky (approved):
I suspect we don’t need it anymore! I presume backend will still understand how to load it without the sha prefix in it?

PR #724, opened 2018-11-26
Don't use componentWillMount in BuildState

ticky:
This was a part of #634, but I figure this is the bit that’s shippable.


plasticine (approved):
2EZ 😁

PR #725, opened 2018-11-26
Unpin dependencies & Update

ticky:
Now that Renovate has been pulled, we’ll need to go back to doing this the manual way.


ticky:
This breaks everything? Oh dear

PR #726, opened 2018-11-26
Add a route we can iterate on the new pipeline things under

plasticine:
:tin:


plasticine:
Will merge when green

keithpitt:
🥫 < correct emoji

PR #727, opened 2018-11-27
React 17 Prep

ticky:
This preps us for React 17 by re-enabling the `react/no-deprecated` lint, and moving the remaining usages of `componentWill{Mount,ReceiveProps}` to their `UNSAFE_` counterparts, which will survive the upgrade to React 17 (for now!)

It also adds lints to warn us against using the UNSAFE_ methods, and we should look at porting away over time.


keithpitt:
@ticky is this mostly shippable now?

ticky:
This is very shippable, it should be a no-op, I just need some eyes on it!

keithpitt (approved):

PR #736, opened 2018-12-10
Try out blueimp-load-image for image resizing

ticky:
This is a proof-of-concept for making images resize on upload. It:

- Checks the image type and size, and
- If the image is big, loads it and resizes it
- If the resized image is smaller than the original, uploads the resized version

### DIFF
--- a/app/components/shared/ImageUploadField.js
+++ b/app/components/shared/ImageUploadField.js
@@ -216,8 +216,10 @@ export default class ImageUploadField extends React.PureComponent<Props, State> 
         if (imageFiles.length === 1) {
           const imageFile = imageFiles[0];
 
-          // If the image is less than 256 kb, or is a gif less than 768kb, don't even bother resizing
-          if (imageFile.size <= 262144 || (imageFile.type === 'image/gif' && imageFile.size <= 786432)) {
+          // If the image is less than 256 kb, is an svg, or is a gif less than 768kb, don't even bother resizing
+          if (imageFile.size <= 262144 ||
+              imageFile.type === 'image/svg' ||
+              (imageFile.type === 'image/gif' && imageFile.size <= 786432)) {
             this.processUpload(imageFile);
             return;
           }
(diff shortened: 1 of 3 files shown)


app/components/shared/ImageUploadField.js
```diff
@@ -3,6 +3,7 @@
 import React from 'react';
 import classNames from 'classnames';
 import styled from 'styled-components';
+import loadImage from 'blueimp-load-image';
```
toolmantim: Before we go live into prod maybe we move this into a dynamic require / bundle. We could even load it only when the first drag+drop starts, and don't finish the existing "uploading" animation until it's loaded, so it's mostly invisible?

ticky: Done! We’ll now load it when you first drag over the document or drag zone (as sometimes you can skip the document! Oops!) and we set the “uploading” animation early.

We also start lazy-loading when you click the “upload” buttons, which means it loads while you’re browsing! Super schmick! 👌🏼

toolmantim: That's really awesome! Nice work!!

app/components/shared/ImageUploadField.js
```diff
@@ -213,37 +238,79 @@ export default class ImageUploadField extends React.PureComponent<Props, State>
...
+                  );
+                } else {
+                  // If loadImage gives us an <img/>, or otherwise couldn't resize it, fall back
+                  this.processUpload(imageFile);
+                }
+              },
+              {
+                minHeight: 50,
+                minWidth: 50,
+                maxHeight: 500,
```
toolmantim: My only question now is: is 500px enough?? It gives us max 250px on a @2x display, or 166px if a @3x ever comes into existence. I'd say that's probably okay? We could make it 1000px if we wanted to be safe 🤔 What do you reckon? I really don't know.

toolmantim: We should totally up the mins though! I only just noticed they're only 50px.

ticky: GitHub’s icons are only 200px (shown at 100pt) at retina, for reference. Depends where you want to take it in the future! I’d suggest that if we want to make it much bigger than 500px we might need to consider some sort of remote image resize solution like IIRC AWS offer.

(Please ignore the min values; that just means if you upload a thing between 50 and 500px we won’t scale it up (this resize utility is pretty bad it turns out))

ticky: actually now I look at it the min should probably be, like, `minWidth: 1` tbqh, we don’t ever want to scale upward

ticky:
Added a hard 2 MiB limit, which I realised I totes forgot. It’s probably not generous enough if we’re going to use a bigger than 500px size, but that’s something we can change down the line if we so choose.

PR #739, opened 2018-12-11
Render Relay Modern pages using a QueryRenderer directly

keithpitt:
- [x] TwoFactor
- [x] GraphQLExplorer
- [x] NewTeam


plasticine:
As a bonus I also discovered that the New Team screen was also stuffed. 😢

plasticine:
OK, think we’re good here — going to back out the reverts and then get this in.

PR #744, opened 2018-12-12
Implement a handler for job link clicks within build annotations

ticky:
It’s been a long time coming, and probably should’ve been part of the round of annotation improvements we launched a while back, but _here it is_.

This adds a click handler to annotations, at the top level (so we don’t mess with the html content client-side!), which;

1. Checks for modifier keys
2. Ignores non-anchor elements, or anchor elements with no `href`
3. Checks the URL, ignoring the hash, matches the current URL
4. Checks a job matching the hash exists on the page
5. If we get here, emits a `job:focus` event in the same way the build header does, and the job is expanded and scrolled to!

# 👩🏻‍🍳👌🏼

### DIFF
--- a/app/components/build/AnnotationsList.js
+++ b/app/components/build/AnnotationsList.js
@@ -1,7 +1,6 @@
 // @flow
 
 import React from 'react';
-import PropTypes from 'prop-types';
 import Relay from 'react-relay/classic';
 
 import PusherStore from 'app/stores/PusherStore';
--- a/app/components/build/AnnotationsList.js
+++ b/app/components/build/AnnotationsList.js
@@ -1,30 +1,30 @@
+// @flow
+
 import React from 'react';
 import PropTypes from 'prop-types';
 import Relay from 'react-relay/classic';
 
 import PusherStore from 'app/stores/PusherStore';
 
-class AnnnotationsList extends React.Component {
-  static propTypes = {
-    build: PropTypes.shape({
-      id: PropTypes.string.isRequired,
-      annotations: PropTypes.shape({
-        edges: PropTypes.arrayOf(
-          PropTypes.shape({
-            node: PropTypes.shape({
-              id: PropTypes.string.isRequired,
-              style: PropTypes.string.isRequired,
-              body: PropTypes.shape({
-                html: PropTypes.string
-              }).isRequired
-            }).isRequired
-          }).isRequired
-        ).isRequired
-      }).isRequired
-    }).isRequired,
-    relay: PropTypes.object
-  };
+type Props = {
+  build: {
+    id: string,
+    annotations: {
+      edges: Array<{
+        node: {
+          id: string,
+          style: string,
+          body: {
+            html: ?string
+          }
+        }
+      }>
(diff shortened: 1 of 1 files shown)


sj26:
Omg yes please amazing!!!

app/components/build/AnnotationsList.js
```diff
@@ -43,6 +42,49 @@ class AnnnotationsList extends React.Component {
...
+      return;
+    }
+
+    // Don't change clicks in things which aren't <a /> elements, or don't have a `href`
+    if (!(event.target instanceof HTMLAnchorElement) || !event.target.href) {
+      return;
+    }
+
+    // Parse the link URL
+    const linkUrl = new URL(event.target.href);
```
keithpitt: Any chance this can raise an error?

keithpitt (approved):

app/components/build/AnnotationsList.js
ticky: If event.target.href is not a valid url, yeah, it’ll raise a TypeError. I guess if it’s not valid this should just, what, continue as normal?

ticky: I can’t actually find a way to provide an “invalid” URL to this, given that all the URLs are normalised through both the annotation gear _and_ the browser.

If an error _does_ get thrown in this handler, the browser will navigate as normal (and the user will likely get a 404, lol), and we might even get it in Bugsnag, if we’re lucky. I’d love to know how it’s even possible!

ticky: I’ve also added an extra layer of insurance against this in b8720de; it provides the current URL as the base to the URL constructor, so even if the browser’s normalisation doesn’t mean we get a fully-qualified URL from the DOM (unlikely, but older browsers didn’t do this right), we’ll always produce what the browser interprets it as, which means domain-relative URLs will always be valid.

keithpitt: Perfect! I don't even think it's a serious issue, just something I thought of.

keithpitt:
This great, peeps are gonna love it!

ticky:
📆🚢

plasticine:
@ticky Sorry to be a pain, but I’m not sure what this does 🙃 Is it just that you can click links in an annotation? Sorry, clearly I’m missing some context on what this change actually is.

keithpitt:
(I'm driving by this PR so I'll hop in)

Here's an annotation with an RSpec failure:

image

And the link's HTML is this:

```html
<a href="https://buildkite.com/buildkite/buildkite/builds/20574#c9f62f17-fc09-4a33-a2db-e5db079dfdbb">Job #c9f62f17-fc09-4a33-a2db-e5db079dfdbb</a>
```

But it doesn't actually take you to the job or expand it. This change just makes it so if you click on a link in an annotation that looks like a reference to a job, we'll do the #righthing.

plasticine:
@keithpitt Ahhhh, yeah I see — very cool! Thanks!

toolmantim (approved):
🙌🏼🚀

PR #745, opened 2018-12-17
JS graph/relay UI tests

plasticine:
screen shot 2018-12-17 at 1 50 31 pm

This was a bit of a late-last-week and bit-of-weekend hack that I demoed a while back and only just finished up.

This PR plumbs things together so that we can write pretty nice, succinct UI units tests against mocked out GraphQL responses. Unlike traditional static mocks or fixtures that can get out of data, this mocking is performed in a pretty unique and awesome way that leverages the fact that the GraphQL schema is vendored in the project, and is typed. This means that mocks will always match the blessed shape of the backend. It also adds a few pretty basic proof-of-concept tests.

Check out [the implementation of the mocking test helper](https://github.com/buildkite/frontend/compare/feature/integration-tests?expand=1#diff-6549614071e526bc61ad1487a3e12a28R1), and an [example of it in use](https://github.com/buildkite/frontend/compare/feature/integration-tests?expand=1#diff-242bea77434b7166bf4e300a2124740eR142
) to get an idea of how this can be used. 

I can’t see us writing heaps of these tests, but I think this still going to be an incredibly useful tool for us to avoid having to lean too heavily on expensive and fragile browser tests.


jest/fileTransformer.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable */
+
+ const path = require('path');
+
+ module.exports = {
+  process(src, filename, _config, _options) {
+    return 'module.exports = ' + JSON.stringify(path.basename(filename)) + ';';
+  }
+};
```
ticky: Is this like, fake Webpack-file-loader? :P

app/components/OrganizationShow/__tests__/OrganizationShow_test.js
```diff
@@ -0,0 +1,152 @@
...
+        ]
+      }
+    }
+  };
+
+  beforeAll(() => {
+    Environment.create();
+    const query = OrganizationShow.query.modern().text;
+    const variables = {organizationSlug: 'some-org', pageSize: 30};
+    graphMock = GraphMock.create(query, variables, graphMockTypes);
```
ticky: yeah this API seems nice ✌🏼

ticky (commented):
Cool Technology!

jest/fileTransformer.js
plasticine: Yeah, it just used as a replacement for the modules as mapped in https://github.com/buildkite/frontend/pull/745/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R193

PR #754, opened 2018-12-19
Add public field to preload

plasticine:
🥫


plasticine:
Going tp FYI this into `master`

PR #764, opened 2019-01-14
Remove nested animation-spin in the spinner

toolmantim:
Looks like we have a double animation happening inside our spinner, which goes against the performance comment at the top of the file. This removes the nested `animation-spin` and fixes up the other uses of the class to work with the new values.


toolmantim:
It'd be nice to roll out this optimisation (pushing the animation to outside the `svg` element) to the `BuildState` icon too…

toolmantim:
I think this is good to go!

PR #767, opened 2019-01-17
Add a "Prettify" button to the in-app GraphQL Console

keithpitt:
Does what is says on the Tin!

image

GraphiQL (the editor that this is loosely based on) uses a pretty basic way of doing "prettification"

```js
print(parse(query))
```

Essentially it parses the query into an AST, and re-prints the AST with some standard formatting rules. The biggest drawback with this approach is that comments aren't part of the AST, so they're dropped on the floor after hitting the "pretty" button - which is kinda shitty.

I've opted for pulling in the "prettier" package instead and which does do the right thing with comments. I'm doing a magical Webpack import thing since it's only used in this one place and there's no need to import for everything and bloat the main packages - hopefully I did it right!

### DIFF
--- a/app/components/user/graphql/GraphQLExplorerConsole.js
+++ b/app/components/user/graphql/GraphQLExplorerConsole.js
@@ -10,7 +10,7 @@ import Dropdown from 'app/components/shared/Dropdown';
 import FlashesStore from 'app/stores/FlashesStore';
 import GraphQLExplorerConsoleEditor from "./GraphQLExplorerConsoleEditor";
 import GraphQLExplorerConsoleResultsViewer from "./GraphQLExplorerConsoleResultsViewer";
-import { executeQuery } from "./query";
+import { executeQuery, prettifyQuery } from "./query";
 import consoleState from "./consoleState";
 import type { RelayProp } from 'react-relay';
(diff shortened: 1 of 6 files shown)


ticky:
Wow, I was going to ask “how big is Prettier?” and yikes!

image

Good call lazy-loading.

app/components/user/graphql/GraphQLExplorerConsole.js
```diff
@@ -282,6 +291,12 @@ class GraphQLExplorerConsole extends React.PureComponent<Props, State> {
     this.executeCurrentQuery();
   };
 
+  handlePrettifyClick = () => {
+    prettifyQuery(this.state.query, (query) => {
+      this.setState({ query: query });
+    });
+  }
```
ticky: I wonder if we’d want to do the lazy-load-on-hover trick here? I did it in ImageUploadField:

https://github.com/buildkite/frontend/blob/bb8aaed01cb7b16aa2aaff4d6ffec8280daf9891/app/components/shared/ImageUploadField.js#L205-L219

ticky (approved):

plasticine:
brb implementing prettier in dev... 😉😜

PR #774, opened 2019-01-21
Flow++, part 1: emit PropTypes based on Flow types

ticky:
~As part of doing some dependency updates, I found a tool to convert PropTypes to Flow annotations, and a Babel plugin to emit PropTypes based on Flow annotations, so this might be a nice way to reduce the burden of switching to Flow. This is just me experimenting with using that.~


plasticine:
Interesting! It’s unavoidable of course but the number of `any`s is a shame.

ticky:
An `any` is at least a good first step on the road to being more specific! 😃

ticky:
I’ve reduced this to just the bit that emits PropTypes based on Flow types. This means we get runtime checks for things called from legacy CoffeeScript, without having to write two sets of types, and that we also get Flow type checks as part of the build.

I’m going to put together PRs for auto-converting some of our PropTypes to Flow types separately.

PR #775, opened 2019-01-21
Fix link to license

alyssais:


ticky:
Looks like _someone_ was authoring this on a case-insensitive, but case-preserving OS 😅 (and never checked it on GitHub)

Thanks for the fix!

PR #778, opened 2019-01-22
Make agent guide buttons black and not lime

toolmantim:


toolmantim:
Closing in favour of a bigger PR that makes the list link styles consistent

PR #779, opened 2019-01-22
Cleanup index link styles

toolmantim:
Now we've updated the link colours, the lists full of green seem a bit harsh. This cleans up the agent/teams/pipelines/members list links, and makes them consistently styled with black/lime-hover.

It also links to the users from the Team’s Members list, to make it consistent with the Team’s Pipelines list.

## Before

image

image

before-2

## After

image

image

after-2


toolmantim:
The schema.json diff is quite huge 🤔

ticky:
> The schema.json diff is quite huge 🤔

Yep, but it looks like it’s mostly whitespace though. I really wish yajl would make up its mind on what format empty arrays should use!

PR #783, opened 2019-01-30
Use standard “require” naming for our require shim

ticky:


ticky:
I believe Webpack would import _every possible import_ if we did it that way, unfortunately! I’d rather not rock the boat too hard as the next few steps of this will remove the existence of this shim altogether!

`window["require"]` won’t ever conflict because the Webpack `require` function gets changed to `__webpack_require__` in dev, and further to a single-character name in production builds. It also can be referenced specifically (though I don’t know why we would) from Webpack-powered code by referencing it specifically as `window["require”]`.

PR #784, opened 2019-01-30
Show more annotations button

eleanorakh:
We currently only display the first 5 annotations for any build. I've used the `ShowMoreFooter` component for consistency, but I'm not sure if I'm happy with how it fits into the page here. Thoughts?
screen shot 2019-01-31 at 11 34 48 am

Here's the gist I used for >5 annotations.
https://gist.github.com/eleanorakh/ed0ccde127431e46e92d29b93b0fe44a

### DIFF
--- a/app/components/build/AnnotationsList.js
+++ b/app/components/build/AnnotationsList.js
@@ -35,7 +35,12 @@ type State = {
   loadingMore: boolean
 };
 
-class AnnnotationsList extends React.Component<Props> {
+class AnnnotationsList extends React.Component<Props, State> {
+
+  state = {
+    loadingMore: false
+  };
+
   componentDidMount() {
     PusherStore.on("build:annotations_change", this.handleWebsocketEvent);
     CentrifugeStore.on("build:annotations_change", this.handleWebsocketEvent);
@@ -64,6 +69,7 @@ class AnnnotationsList extends React.Component<Props> {
       return (
         <ShowMoreFooter
           connection={this.props.build.annotations}
+          loading={this.state.loadingMore}
           onShowMore={this.handleShowMoreAnnotations}
         />
       );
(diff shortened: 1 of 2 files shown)


app/components/build/AnnotationsList.js
```diff
@@ -41,10 +57,40 @@ class AnnnotationsList extends React.Component<Props> {
...
-      <div>{annotations}</div>
+      <div>
+        {annotations}
+        {this.renderShowMore()}
+      </div>
     );
   }
 
+  renderShowMore() {
+    if (this.props.build.annotations.pageInfo.hasNextPage) {
```
ticky: You don’t need to check for hasNextPage here; `ShowMoreFooter` can check that for you:

https://github.com/buildkite/frontend/blob/e8d97f5d5c9ed2faba3f8924894b02d1c59e9016/app/components/shared/ShowMoreFooter/index.js#L26-L38

app/components/build/AnnotationsList.js
```diff
@@ -162,7 +212,13 @@ export default Relay.createContainer(AnnnotationsList, {
                 html
               }
             }
+            cursor
+          }
+          pageInfo {
+            hasNextPage
+            hasPreviousPage
```
ticky: If we rely on `ShowMoreFooter` to handle the paging info, we don’t need to pull in either of these into this component!

app/components/build/AnnotationsList.js
eleanorakh: Oh awesome!

app/components/build/AnnotationsList.js
eleanorakh: That's very cool, thanks for pointing it out :)

eleanorakh:
@ticky I have a feeling that https://github.com/buildkite/buildkite/pull/3468 needs to merge first once I figure out what's up with the specs 🤔

